PR #41, opened 2022-04-08
Add time travel arrows

krsjoseph:
[Notion Task](https://www.notion.so/lncapital/Arrows-to-quickly-jump-timeframes-33ccbaab3d8c404aab0c1335a6cfd59b)

https://user-images.githubusercontent.com/1019869/162409628-800a0c7a-5843-4702-bc9e-874730f40f97.mp4

### DIFF
--- a/web/src/components/timeIntervalSelect/interval_select.scss
+++ b/web/src/components/timeIntervalSelect/interval_select.scss
@@ -185,6 +185,9 @@
     // TODO: Quick hack, clean up. FLW
     margin-top: -2px;
   }
+  .interval {
+    min-width: 195px;
+  }
   .interval .previous {
     color: var(--fg-muted);
   }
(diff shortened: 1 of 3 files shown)


web/src/components/timeIntervalSelect/interval_select.scss
```diff
@@ -168,7 +168,7 @@
   }
 }
 
-.torq .time-interval-wrapper {
+.torq #time-interval-wrapper {
```
ohenrik, line 171: is `time-interval-wrapper` applied as both an id and a class? Or where the css bellow not applied before?

ohenrik:
This looks good!

web/src/components/timeIntervalSelect/interval_select.scss
krsjoseph: I changed this to an ID to have more specificity  and not use any `!important` 's

krsjoseph:
Changes addressed @ohenrik (y)

PR #57, opened 2022-08-18
new payment endpoint + tests

iulian-moraru:


ohenrik:
Great work on this. I ended up switching to websockets communication to capture updated status for payments, hence the large change.

PR #60, opened 2022-08-19
new address & send payment to address

iulian-moraru:
1. new address endpoint: api/on-chain-tx/newaddress   
POST method
   input fields:
      - type int = address type:
      //AddressType has to be one of:
      	//- `p2wkh`: Pay to witness key hash (`WITNESS_PUBKEY_HASH` = 0)
      	//- `np2wkh`: Pay to nested witness key hash (`NESTED_PUBKEY_HASH` = 1)
      	//- `p2tr`: Pay to taproot pubkey (`TAPROOT_PUBKEY` = 4)
      - account string = name of the account to generate a new address for. If empty, the default wallet account is used.

2. send coins endpoint: api/on-chain-tx/sendcoins
   POST METHOD
   input fields:
      - addr string = The address to send coins to
      - amoint int = The amount in satoshis to send
      - satpervbyte *uint64 = manual fee rate set in sat/vbyte that should be used when crafting the transaction


internal/on_chain_tx/new_address.go
```diff
@@ -0,0 +1,25 @@
+package on_chain_tx
+
+import (
+	"context"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/rs/zerolog/log"
+)
+
+func newAddress(client lnrpc.LightningClient, addressType int32, account string) (r string, err error) {
```
ohenrik, line 9: You should create a new interface that only requires the function you need (client.NewAddress)

That way you do not have to mock every other function in the tests.

internal/on_chain_tx/new_address_test.go
```diff
@@ -0,0 +1,352 @@
...
+	"context"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"google.golang.org/grpc"
+	"testing"
+)
+
+type MockNewAddressLClnt struct {
+}
+
+func (m MockNewAddressLClnt) WalletBalance(ctx context.Context, in *lnrpc.WalletBalanceRequest, opts ...grpc.CallOption) (*lnrpc.WalletBalanceResponse, error) {
```
ohenrik, line 13: Here, all mock implementations but `NewAddress` can be removed.

internal/on_chain_tx/new_address_test.go
```diff
@@ -0,0 +1,352 @@
...
+	//TODO implement me
+	panic("implement me")
+}
+
+func (m MockNewAddressLClnt) SubscribeCustomMessages(ctx context.Context, in *lnrpc.SubscribeCustomMessagesRequest, opts ...grpc.CallOption) (lnrpc.Lightning_SubscribeCustomMessagesClient, error) {
+	//TODO implement me
+	panic("implement me")
+}
+
+func (m MockNewAddressLClnt) NewAddress(ctx context.Context, in *lnrpc.NewAddressRequest, opts ...grpc.CallOption) (*lnrpc.NewAddressResponse, error) {
```
ohenrik, line 333: All except this one.

internal/on_chain_tx/send_coins.go
```diff
@@ -0,0 +1,43 @@
+package on_chain_tx
+
+import (
+	"context"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/rs/zerolog/log"
+	"google.golang.org/grpc"
+)
+
+type lndClientSendCoins interface {
```
ohenrik, line 10: This is the correct way to implement it.

internal/on_chain_tx/send_coins_test.go
```diff
@@ -0,0 +1,33 @@
...
+func TestSendCoins(t *testing.T) {
+	address := "test"
+	amount := 123
+	var satPerVbyte *uint64
+
+	client := MockSendCoinsLClnt{}
+	resp, err := sendCoins(client, address, int64(amount), satPerVbyte)
+
+	if err != nil {
+		t.Fail()
```
ohenrik, line 27: You should return the error here.

PR #66, opened 2022-09-07
Add command "start-dev-flow" to start nodes activity

iulian-moraru:


Makefile
```diff
@@ -74,3 +77,7 @@ stop-dev-env:
 purge-dev-env:
 	go build ./virtual_network/torq_vn &&  go run ./virtual_network/torq_vn purge --db true
 
+# Start flow
+.PHONY: start-dev-flow
+start-dev-flow:
+	go build ./virtual_network/torq_vn && go run ./virtual_network/torq_vn flow --invfrq $(invfrq) --scofrq $(scofrq) --ochfrq $(ochfrq)
```
maxwedwards, line 83: go run will build and run from source but won't leave a torq_vn binary around after. So no need for the go build step at the start.

Makefile
```diff
@@ -3,6 +3,9 @@ testDbPort = 5433
 backendTest = go test ./... -v -count=1
 frontendTest = cd web && npm i && npm test -- --watchAll=false
 stopDevDb = ($(MAKE) stop-dev-db && false)
+invfrq = 1
+scofrq = 30
+ochfrq = 10
```
maxwedwards, line 8: Got a nit about these names. I couldn't tell what they were without looking. In the make file can we call the vars something like: `virtualNetworkInvoiceFrequency` etc etc. Also could we change the flags in torq_vn to: `--invoice-frequency`. Longer but it's unlikely we will be typing this out much and I think they tell you what they are.

virtual_network/torq_vn/torq_vn.go
```diff
@@ -117,6 +117,46 @@ func main() {
...
+			&cli.IntFlag{
+				Name:  "invfrq",
+				Usage: "Set invoice creation frequency - seconds",
+			},
+			&cli.IntFlag{
+				Name:  "scofrq",
+				Usage: "Create address and send coins frequency - seconds",
+			},
+			&cli.IntFlag{
+				Name:  "ochfrq",
```
maxwedwards, line 142: As comment above

maxwedwards (changes requested):
Nice! Few nits around variable naming in the makefile and flags for the main program.

Main thing though would be can we drop all the time.sleeps please? I'd like to keep us with no sleeps in this virtual network and in tests (unless it's a wait before a retry).

PR #67, opened 2022-09-14
endpoints requests based on local node ID

iulian-moraru:


internal/settings/node_connection_details.go
```diff
@@ -14,49 +14,40 @@ type ConnectionDetails struct {
 	Deleted           bool
 }
 
-func GetConnectionDetails(db *sqlx.DB, allNodes bool, nodeId int) ([]ConnectionDetails, error) {
-	var localNodes []localNode
-	var activeNode localNode
-	var err error
+func GetActiveNodesConnectionDetails(db *sqlx.DB) (activeNodes []ConnectionDetails, err error) {
```
maxwedwards, line 17: I wasn't really a fan of having one function that behaved so differently depending on the input so I split it into two functions. One that gets the specific id and one that gets all non deleted and non disabled.

internal/channels/batch_open_channel.go
```diff
@@ -14,28 +14,20 @@ import (
 func batchOpenChannels(db *sqlx.DB, req BatchOpenRequest) (r BatchOpenResponse, err error) {
 	bOpenChanReq, err := checkPrepareReq(req)
 	if err != nil {
-		log.Error().Msgf("Error checkPrepareReq: %v", err)
-		return r, err
+		return BatchOpenResponse{}, err
```
maxwedwards, line 17: Our preferred style when returning an error is to explicitly return nil or zero value for the first item. So here I'm explicitly returning a fresh/empty/zero value BatchOpenResponse{} to make it obvious. I do this anywhere we are returning an error.

internal/channels/batch_open_channel.go
```diff
@@ -14,28 +14,20 @@ import (
...
 	}
 
-	connectionDetails, err := settings.GetConnectionDetails(db, false, req.NodeId)
+	connectionDetails, err := settings.GetNodeConnectionDetailsById(db, req.NodeId)
 	if err != nil {
-		log.Error().Err(err).Msgf("Error getting node connection details from the db: %s", err.Error())
-		return r, errors.New("Error getting node connection details from the db")
```
maxwedwards, line 24: Few things to note here:

Either return an error OR log. If you return an error it's then up the the calling function to log it or bubble it up further.
When using the Err(err) function on zerolog there is no need to add the error to the message string
If you have an error you should generally wrap it and bubble it up rather than creating a new error because otherwise you lose all the wrapped error history and context
And as already mentioned, don't return r variable when erroring but instead nil or zero value item

internal/channels/batch_open_channel.go
```diff
@@ -14,28 +14,20 @@ import (
...
-		connectionDetails[0].GRPCAddress,
-		connectionDetails[0].TLSFileBytes,
-		connectionDetails[0].MacaroonFileBytes)
+		connectionDetails.GRPCAddress,
+		connectionDetails.TLSFileBytes,
+		connectionDetails.MacaroonFileBytes)
 	if err != nil {
-		log.Error().Err(err).Msgf("can't connect to LND: %s", err.Error())
-		return r, errors.Newf("can't connect to LND")
+		return BatchOpenResponse{}, errors.Wrap(err, "Connecting to LND")
```
maxwedwards, line 30: Here you can see me instead wrapping the error and returning a zero value BatchOpenResponse.

Also note, when creating and wrapping errors, avoid terms such as "can't do x, error with x, problem doing x" etc etc and just say what was trying to happen. So rather than "Can't connect to LND" it's now "Connecting to LND".

The one time you can say "Problem with" or "Error doing" is at the top level when you log the error out. I'll write a guide on error handling to put in out contributors.md

internal/payments/new_payment.go
```diff
@@ -108,21 +107,16 @@ func SendNewPayment(
...
 	}
 	conn, err := lnd_connect.Connect(
-		connectionDetails[0].GRPCAddress,
-		connectionDetails[0].TLSFileBytes,
-		connectionDetails[0].MacaroonFileBytes)
+		connectionDetails.GRPCAddress,
+		connectionDetails.TLSFileBytes,
+		connectionDetails.MacaroonFileBytes)
 	if err != nil {
-		server_errors.WrapLogAndSendServerError(c, err, "Failed connecting to LND")
```
maxwedwards, line 116: server_errors is designed to work with REST Handlers only and not with websockets. Just as a note when calling server_errors, you would need a return statement after. I replaced it with an errors.Wrap().

internal/payments/new_payment.go
```diff
@@ -108,21 +107,16 @@ func SendNewPayment(
...
 
-	connectionDetails, err := settings.GetConnectionDetails(db, false, npReq.NodeId)
+	connectionDetails, err := settings.GetNodeConnectionDetailsById(db, npReq.NodeId)
 	if err != nil {
-		return errors.New("Error getting node connection details from the db")
-	}
-
-	if len(connectionDetails) == 0 {
-		//log.Debug().Msgf("Node is deleted or disabled")
-		return errors.Newf("Local node disabled or deleted")
```
maxwedwards: I spotted this a few times, using `Newf` or `Msgf` with no formatting directive. If it's a plain string message then just use `New` or `Msg`

PR #69, opened 2022-09-17
Add static channel backup endpoint

jonassagild:
This adds an endpoint for getting a static channel backup. 

~~A test of the endpoint is missing but will be added.~~ (EDIT: couldn't find a way to test this except on an integration level)

Happy to receive feedback on anything and everything.

### DIFF
--- a/internal/channels/static_channel_backup.go
+++ b/internal/channels/static_channel_backup.go
@@ -1,0 +1,40 @@
+package channels
+
+import (
+	"context"
+
+	"github.com/cockroachdb/errors"
+	"github.com/jmoiron/sqlx"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/lncapital/torq/internal/settings"
+	"github.com/lncapital/torq/pkg/lnd_connect"
+	"github.com/rs/zerolog/log"
+)
+
+func getStaticChannelBackup(db *sqlx.DB, requestBody lnrpc.ChanBackupExportRequest) (r lnrpc.ChannelBackups, err error) {
+
+	connectionDetails, err := settings.GetConnectionDetails(db)
+	conn, err := lnd_connect.Connect(
+		connectionDetails[0].GRPCAddress,
+		connectionDetails[0].TLSFileBytes,
+		connectionDetails[0].MacaroonFileBytes)
+	if err != nil {
+		log.Error().Err(err).Msgf("can't connect to LND: %s", err.Error())
+		return r, errors.Newf("can't connect to LND")
+	}
+
+	defer conn.Close()
+
+	client := lnrpc.NewLightningClient(conn)
+	ctx := context.Background()
+
+	resp, err := client.ExportAllChannelBackups(ctx, &requestBody)
+	if err != nil {
+		log.Error().Msgf("Error exporting all channel backups: %v", err)
+		return r, err
+	}
+
+	staticChannelBackupResp := *resp.SingleChanBackups
+
+	return staticChannelBackupResp, err
+}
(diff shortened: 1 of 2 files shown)


internal/channels/static_channel_backup.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/lncapital/torq/pkg/lnd_connect"
+	"github.com/rs/zerolog/log"
+)
+
+func getStaticChannelBackup(db *sqlx.DB) (r lnrpc.ChannelBackups, err error) {
+	connectionDetails, err := settings.GetConnectionDetails(db)
+	conn, err := lnd_connect.Connect(
+		connectionDetails[0].GRPCAddress,
+		connectionDetails[0].TLSFileBytes,
+		connectionDetails[0].MacaroonFileBytes)
```
maxwedwards, line 19: The api should be told for which node we should connect to rather than hardcoding it to the first one. Probably a query string param.

internal/channels/static_channel_backup.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		log.Error().Err(err).Msgf("can't connect to LND: %s", err.Error())
+		return r, errors.Newf("can't connect to LND")
+	}
```
maxwedwards, line 23: Functions should generally either log the error OR return it to the caller who can decide what to do with it. In this case simply returning it will allow `getStaticChannelBackupHandler` to log the problem. As it is now it will be logged twice.

Error messages ideally should not use phases such as "can't do x" and "error doing x" and instead should simply say _what_ has just failed such as "Connecting to LND node".

Finally, we should probably wrap the error returned from LND_connect.Connect rather than creating a new error

internal/channels/static_channel_backup.go
```diff
@@ -0,0 +1,41 @@
...
+
+	client := lnrpc.NewLightningClient(conn)
+	ctx := context.Background()
+
+	var chanBackupExportReq lnrpc.ChanBackupExportRequest
+
+	resp, err := client.ExportAllChannelBackups(ctx, &chanBackupExportReq)
+	if err != nil {
+		log.Error().Msgf("Error getting static channel backup: %v", err)
+		return r, err
```
maxwedwards, line 35: Again, same comment as above. Either return a (probably wrapped) error or log it, no need for both.

internal/channels/static_channel_backup.go
```diff
@@ -0,0 +1,41 @@
...
+	}
+
+	staticChannelBackupResp := *resp.SingleChanBackups
+
+	return staticChannelBackupResp, err
```
maxwedwards, line 40: I'd say it would be preferred to explicitly return nil here rather than err even though I know err will be nil. Just makes it clearer that we aren't returning an error.

```suggestion
	return staticChannelBackupResp, nil
```

maxwedwards (changes requested):
Great PR, just a few little things to clean up. I realise some of the problems I've mentioned exist in our codebase elsewhere but we are trying to clean these things up as we go :)

jonassagild:
Great feedback, thanks. I will have a look at them shortly!

internal/channels/static_channel_backup.go
jonassagild: Agree!

maxwedwards:
@jonassagild let me know when you are ready for this to be re-reviewed

maxwedwards:
I've marked as draft, just hit "Ready for review" when you want us to take a look

maxwedwards:
Closing for now, if you wish to continue please feel free to reopen.

PR #96, opened 2022-09-24
Display modals from routes

cryptosalomao:
This update allows modals to pop up by changing the current to location to its route. It's going to be useful for performing certain actions without changing the page, but still updating the router location


maxwedwards (approved):
Looks good, just rebase and force push and we can get this merged in.

PR #101, opened 2022-09-27
Router refactor

cryptosalomao:
This is a proposal for a better router component, by using a declarative routing layout and centralizing the routes definitions under a single file


maxwedwards (approved):
Nice looking refactor, much cleaner in this new style. Torq has to run either on root `/` or on a sub path `/torq` which is to support BTCPayServer and this refactor doesn't look it changes that. Merging.

PR #102, opened 2022-09-27
Add prettier

cryptosalomao:
Initial setup for prettier and rules for code style enforcement


web/package.json
```diff
@@ -39,7 +39,9 @@
     "start": "react-scripts start",
     "build": "react-scripts build",
     "test": "react-scripts test",
-    "lint": "eslint --ignore-path .eslintignore --ext .js,.ts,.tsx,.jsx .",
+    "lint": "eslint --ignore-path .eslintignore --fix --ext .js,.ts,.tsx,.jsx .",
```
maxwedwards, line 42: Linting is part of `make test` and I would prefer that `make test` does not modify the source code. I think we can achieve the same goal with the format script by having it be `"npm run prettier && npm run lint -- --fix"`

web/package.json
```diff
@@ -39,7 +39,9 @@
...
+    "prettier": "prettier --write '**/*.{ts,tsx}'",
+    "format": "npm run prettier && npm run lint",
```
maxwedwards, line 44:
```suggestion
    "format": "npm run prettier && npm run lint -- --fix",
```

maxwedwards (changes requested):
We have been living with just using prettier in the editor but I can see the benefit of having the script installed. Please can we just stick to the prettier defaults though?

cryptosalomao:
As discussed with @ohenrik, I'm going to close this PR for now and work on a new one addressing the target for having prettier working alongside eslint.

PR #122, opened 2022-09-30
Refactor API integration

cryptosalomao:
Proposed improvements:

- [x] **Refactor API URL builder**
We currently have a lot of imperative logic exposed into the context of the API slice. With the proposed URL builder in this PR, we wrap the details on the base URL mounting process, keeping the desired behavior to work seamlessly both when running the application directly or when embedded in a third-party platform.
- [x] **Refactor main API slice**
The current API slice still has a lot of hard coded API routes. This PR will also introduce a change to that, so it becomes easier to maintain the application functionalities when changes on routes are needed.
- [x] **Create URL search parameters builder**
The query parameters are currently embedded directly into each of the endpoints inside the slice definition as template strings. A new functionality to build those parameters without hard coding them is going to be shipped within this PR as well.


web/src/constants/third-party-integration.ts
```diff
@@ -0,0 +1 @@
+export const BASE_PATHNAME_PREFIX = "/torq";
```
maxwedwards, line 1: Naming wise I'm not sure "third party integration" is the correct term as yes currently this is used on BTCPayServer but there's nothing stopping someone from running Torq on their own hostname under the sub path torq. Perhaps "subpath-support.ts" might be better as it doesn't assume why someone would use it.

web/src/utils/apiUrlBuilder.ts
```diff
@@ -0,0 +1,28 @@
+import { PROD_ENVIRONMENT_PORT, REST_API_PATHNAME } from "constants/backend";
+import { BASE_PATHNAME_PREFIX } from "constants/third-party-integration";
+
+const isRunningOnThirdParty = window.location.pathname.startsWith(BASE_PATHNAME_PREFIX);
```
maxwedwards, line 4: same naming thing as above. I think I prefer something like isRunningOnSubpath as it's more precise.

maxwedwards:
Looks sweet so far! Added a few comments on naming.

web/src/constants/third-party-integration.ts
cryptosalomao: I kinda associated this subpath a solely related to when running torq inside another environment (Umbrel, BTCPayServer, etc), but totally agree with your suggestion here.

web/src/utils/apiUrlBuilder.ts
cryptosalomao: Agreed.

cryptosalomao:
This PR was going to introduce more changes to the type definitions on each exported API hook, introducing the type definitions for the returned data, but I realized that it wouldn't be simple to add that feature, because doing so will require more work to be done on the components, so I'll postpone this step for now, since we already have plans for refactoring the components.

Issue #129, opened 2022-10-03
Bug: Nullpointer on payload in forwardsSlice.


2Fast2BCn:
Console shows nullpointer problem.

PR #130, opened 2022-10-03
Fix issue #129

2Fast2BCn:
Fixing console error.


maxwedwards:
Unable to replicate problem.

Doesn't seem right to move from `!==' to '!=' as the check is more imprecise.

Looks like non prettier formatting change?

Issue #92, opened 2022-09-23
Feature Request: Add support for other languages.


2Fast2BCn:
We can start using resource keys so it can be translated to support other languages.

PR #131, opened 2022-10-03
#92: Add support for other languages.

2Fast2BCn:
Add support for other languages.

### DIFF
--- a/cmd/torq/internal/torqsrv/server.go
+++ b/cmd/torq/internal/torqsrv/server.go
@@ -194,6 +194,15 @@ func registerStaticRoutes(r *gin.Engine) {
 			c.File("./web/build/static/" + parts[len(parts)-2] + "/" + parts[len(parts)-1])
 			return
 		}
+
+		// locales json files
+		if strings.Contains(path, "/locales/") && strings.Contains(path, ".") &&
+			(strings.Contains(path, "json")) {
+			parts := strings.Split(path, "/")
+			c.File("./web/build/locales/" + parts[len(parts)-1])
+			return
+		}
+
 		c.File("./web/build/index.html")
 	})
 }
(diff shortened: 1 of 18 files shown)


database/migrations/000041_add_language_to_settings.down.psql
```diff
@@ -0,0 +1 @@
+ALTER TABLE settings DROP COLUMN default_language;
```
maxwedwards, line 1: We don't bother with down migrations as for safety we will never run a down migration

maxwedwards: So we just leave the file blank

web/src/App.test.tsx
```diff
@@ -13,5 +13,9 @@ test("renders login page", () => {
     </BrowserRouter>
   );
 
-  expect(getByText(/login/i)).toBeInTheDocument();
+  expect(getByText(/loading/i)).toBeInTheDocument();
+
+  waitFor(() => {
+    expect(getByText(/login/i)).toBeInTheDocument()
+  });
```
maxwedwards, line 20: Any need for the waitfor? That's the default behaviour anyway.

cmd/torq/internal/torqsrv/server.go
```diff
@@ -194,6 +194,15 @@ func registerStaticRoutes(r *gin.Engine) {
 			c.File("./web/build/static/" + parts[len(parts)-2] + "/" + parts[len(parts)-1])
 			return
 		}
+
+		// locales json files
+		if strings.Contains(path, "/locales/") && strings.Contains(path, ".") &&
+			(strings.Contains(path, "json")) {
```
maxwedwards, line 200: Combine the "." and the "json" to be ".json"?

web/src/App.test.tsx
2Fast2BCn: I tried but it does not work without it :-/

Issue #132, opened 2022-10-03
Feature Request: Create summary header section on Payments page.


2Fast2BCn:
Create summary header section on Payments page above the table.

maxwedwards:
This issue has been marked stale because it has been open 60 days with no activity. Remove stale label or comment or this will be closed in 7 days.

PR #133, opened 2022-10-03
#132: Create summary header section on Payments page.

2Fast2BCn:


internal/summary/handlers.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	r, err := getPaymentsSummary(db, filter)
+	switch err.(type) {
+	case nil:
+		break
+	default:
+		server_errors.LogAndSendServerError(c, err)
+		return
+	}
```
maxwedwards, line 48: Simple `if err != nil` should work here right?

If you wanted to check the actual type of error, use `errors.Is` (checks value of error) and `errors.As` (checks type of error).

internal/summary/payments.go
```diff
@@ -0,0 +1,51 @@
...
+			`).
+		PlaceholderFormat(sq.Dollar).
+		From("payment").
+		Where(filter).
+		Prefix(`WITH
+			tz AS (select preferred_timezone as tz from settings),
+			pub_keys as (select array_agg(pub_key) from local_node)
+		`)
+
+	qs, args, err := qb.ToSql()
```
maxwedwards, line 31: Just use a multiline golang string raw quote? It's what we have tended to do.

internal/summary/payments.go
```diff
@@ -0,0 +1,51 @@
...
+		&r.TotalAmt,
+		&r.TotalFees,
+	)
+
+	switch err {
+	case nil:
+		break
+	default:
+		return nil, err
+	}
```
maxwedwards, line 48: Same as above, `if err != nil` should suffice

web/src/apiSlice.ts
```diff
@@ -68,12 +68,18 @@ export const torqApi = createApi({
...
-    >({
+      >({
       query: ({ limit, offset, order, filter }) =>
         `payments?limit=${limit || 100}&offset=${offset || 0}${order ? "&order=" + JSON.stringify(order) : ""}${
           filter ? "&filter=" + JSON.stringify(filter) : ""
         }`,
     }),
+    getPaymentsSummary: builder.query<
+      any,
+      { filter?: any }
```
maxwedwards, line 79: We are trying to move away from using `any`s and always using specific types or `unknown` if it is truly unknown. In this case I think we know the type.

web/src/features/templates/tablePageTemplate/table-page-template.module.scss
```diff
@@ -50,6 +50,57 @@
   grid-area: Totals;
   display: grid;
 }
+
+.pageRow {
+  display: grid;
```
maxwedwards, line 55: Due to some performance problems in safari we have been moving to using flexbox over grid. I'm not sure it matters much here but seems to matter a lot more when there are a lot of them such as in a table of data.

Thoughts @ohenrik ?

web/src/features/transact/Payments/PaymentsHeader.tsx
```diff
@@ -0,0 +1,41 @@
+import styles from "../../templates/tablePageTemplate/table-page-template.module.scss";
```
maxwedwards, line 1: Are these styles really going to be used anywhere else or should they be in a module.scss file local and specific to this file?

web/src/features/transact/Payments/PaymentsHeader.tsx
```diff
@@ -0,0 +1,41 @@
...
+import classNames from "classnames";
+import * as d3 from "d3";
+import {useGetPaymentsSummaryQuery} from "../../../apiSlice";
+import {deserialiseQuery} from "../../sidebar/sections/filter/filter";
+import {useAppSelector} from "../../../store/hooks";
```
maxwedwards, line 6: The way we have setup our frontend build is that you don't need to specify relative paths like this. We told node to look in the `web/src` folder for packages. This means that `import {useGetPaymentsSummaryQuery} from "apiSlice";` would work.

web/src/features/transact/Payments/PaymentsPage.tsx
```diff
@@ -269,10 +270,16 @@ function PaymentsPage() {
       offsetHandler={setOffset}
     />
   );
+
+  const pageTotals = (
```
maxwedwards, line 274: uneeded variable? can just be inlined

web/src/features/transact/Payments/PaymentsPage.tsx
```diff
@@ -269,10 +270,16 @@ function PaymentsPage() {
...
+    <PaymentsHeader />
+  );
+
   return (
     <TablePageTemplate
       title={"Payments"}
       breadcrumbs={breadcrumbs}
+      pageTotals={pageTotals}
```
maxwedwards, line 282: as above, just inline?

cmd/torq/internal/torqsrv/server.go
```diff
@@ -144,6 +145,11 @@ func registerRoutes(r *gin.Engine, db *sqlx.DB, apiPwd string, restartLNDSub fun
 			flow.RegisterFlowRoutes(flowRoutes, db)
 		}
 
+		summaryRoutes := api.Group("/summary")
+		{
+			summary.RegisterSummaryRoutes(summaryRoutes, db)
+		}
```
maxwedwards, line 151: Do we really need a new route for this? Can't the data be summarised on the front end from an existing api call with the full data?

2Fast2BCn: I wrote a separate query for performance reasons. I didn't look into the codebase yet on how paging is done but if it returns partial data with offsets then I'm not sure using an existing api call retrieving all data only to show a summary is the best option?

In my opinion the summary should be for all pages not only the visible one.

web/src/features/transact/Payments/PaymentsHeader.tsx
2Fast2BCn: I think this style of table header might be added on other pages too right? If not then it makes sense to move it.

web/src/features/templates/tablePageTemplate/table-page-template.module.scss
2Fast2BCn: I kinda stole the layout from web/src/features/channel/channel-page.module.scss

maxwedwards:
Going to close for now until this comes back with clearer figma design and requirements

PR #134, opened 2022-10-03
Subscribe features

iulian-moraru:


maxwedwards:
Looks like it now has a backend test failing. Are you able to look into that @iulian-moraru?

maxwedwards:
It looks like you've pulled in all sorts of changes into this PR. Could you create a clean feature branch based off updated main and pull in just the changes and create a new PR? On of the the things I think might be confusing things is that this PR is based off your fork's main rather than just a feature branch.

Issue #114, opened 2022-09-29
Add tags to channels


ohenrik:
The backend (API) for this feature is already built but is disabled. It should be possible to tag a channel with user defined tags. These tags can be used to quickly select channels and perform actions on onn of them combined or to group channels.

ohenrik:
Closed by #189

PR #156, opened 2022-10-12
WIP #114: Add tags to channels

2Fast2BCn:


maxwedwards:
I've only had a very quick scan as this is quite big but it's looking quite smart at first glance. Spoke on slack about few nits around typsecript anys, snake_case go files/folders/package names and error handling conventions.

internal/cache/cacheHelper.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"fmt"
+	"github.com/jmoiron/sqlx"
+	"github.com/lncapital/torq/internal/channels"
+	"github.com/lncapital/torq/internal/corridors"
+	"github.com/lncapital/torq/internal/peers"
+)
+
+func ReinitializeCache(db *sqlx.DB) {
+	fmt.Println("Loading caches in memory.")
```
maxwedwards, line 12: Use zero log for all logging. Perhaps this is `Info()` level?

internal/cache/cacheHelper.go
```diff
@@ -0,0 +1,16 @@
...
+	corridors.RefreshCorridorCache(db)
+	peers.RefreshPeersCache(db)
+	channels.RefreshChannelsCache(db)
```
maxwedwards, line 15: Not handling errors?

internal/channel_history/channel_info.go
```diff
@@ -8,7 +8,8 @@ import (
 
 type channel struct {
 	// Node Alias
-	Alias null.String `json:"alias"`
+	Alias  null.String `json:"alias"`
+	PeerId null.String `json:"peerId"`
```
maxwedwards, line 12: Not asking for a change as I see it matches the existing types but I've started just using pointer of string for nullable strings rather than null.String as they seem to work ok with both json parsers and sqlx.

internal/channel_tags/channelTag.go
```diff
@@ -0,0 +1,22 @@
...
+	ChannelId    int       `json:"channelId" db:"channel_db_id"`
+	TagId        int       `json:"tagId" db:"tag_id"`
+	CreatedOn    time.Time `json:"createdOn" db:"created_on"`
+	// No UpdateOn as there will never be an update always create/delete.
+}
+
+type TagOrigin int
+
+const (
+	Corridor = TagOrigin(iota)
```
maxwedwards, line 21: Nice

maxwedwards, line 21: Do these need to be exported? From my quick search they are only used in the channel_tag package

internal/channel_tags/database.go
```diff
@@ -0,0 +1,81 @@
...
+	"github.com/lncapital/torq/internal/channels"
+	"github.com/lncapital/torq/internal/corridors"
+	"github.com/lncapital/torq/internal/database"
+	"github.com/lncapital/torq/internal/tags"
+	"github.com/pkg/errors"
+	"time"
+)
+
+func getChannelTag(db *sqlx.DB, channelTagId int) (channelTag, error) {
+	ct := &channelTag{}
```
maxwedwards, line 15: making ct a pointer and then dereferencing it on the return from this func, remove the & from this line and the * from the return and then no need to use a pointer at all.

internal/channel_tags/database.go
```diff
@@ -0,0 +1,81 @@
...
+	err := db.Get(&ct, "SELECT * FROM channel_tag WHERE channel_tag_id = $1;", channelTagId)
+	if err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return channelTag{}, nil
+		}
+		return channelTag{}, errors.Wrap(err, database.SqlExecutionError)
+	}
+	return *ct, nil
```
maxwedwards, line 23: As above, remove *

internal/channel_tags/database.go
```diff
@@ -0,0 +1,81 @@
...
+}
+
+func addChannelTag(db *sqlx.DB, channelTag channelTag) (*channelTag, error) {
```
maxwedwards, line 26: Does this need to be a pointer to channelTag or can it just be by value?

internal/channel_tags/database.go
```diff
@@ -0,0 +1,81 @@
...
+		return nil, errors.Wrap(err, database.SqlExecutionError)
+	}
+
+	return &channelTag, nil
+}
+
+func GenerateChannelTag(db *sqlx.DB) error {
+	channels, err := channels.GetAllChannels(db)
+	if err != nil {
+		return errors.Wrap(err, "Unable to obtain channels for channel tag generation.")
```
maxwedwards, line 49: Just state what was attempted without prefixing with "unable" or postfixing with how it may have been called. "Getting all channels" should be sufficient to give context to what was just attempted.

internal/channel_tags/database.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	tags, err := tags.GetTags(db)
+	if err != nil {
+		return errors.Wrap(err, "Unable to obtain tags for channel tag generation.")
```
maxwedwards, line 53: "Getting tags" should be a sufficient error context

internal/channel_tags/database.go
```diff
@@ -0,0 +1,81 @@
...
+	}
+	tx, err := db.Begin()
+	if err != nil {
+		return errors.Wrap(err, database.SqlExecutionError)
```
maxwedwards, line 57: I think this is too generic a wrap. You can be more specific and state that it was attempting to start a transaction.

internal/channel_tags/database.go
```diff
@@ -0,0 +1,81 @@
...
+				_, err = tx.Exec(`
+					INSERT INTO channel_tag(node_id,peer_id,channel_db_id,tag_origin_id,tag_id,created_on)
+					VALUES ($1, $2, $3, $4, $5, $6);`,
+					channel.LocalNodeId, channel.PeerId, channel.ChannelDBID, Corridor, tag.TagId, time.Now().UTC())
+				if err != nil {
+					return errors.Wrap(err, database.SqlExecutionError)
+				}
+			}
+		}
+	}
```
maxwedwards, line 75: Shouldn't there be a `defer tx.Rollback()` just after opening the transaction? If there is an error anywhere in this code the transaction won't be rolled back will it?

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	"github.com/lncapital/torq/internal/corridors"
+	"github.com/lncapital/torq/pkg/server_errors"
+	"gopkg.in/guregu/null.v4"
+	"log"
+	"net/http"
+	"strconv"
+)
+
+func RegisterChannelTagRoutes(r *gin.RouterGroup, db *sqlx.DB) {
+	r.POST("add", func(c *gin.Context) { addChannelTagHandler(c, db) })
```
maxwedwards, line 17: Do we need the route `api/tags/add`? Isn't a POST to `/api/tags/` cleaner?

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	r.DELETE(":channelTagId", func(c *gin.Context) { removeChannelTagHandler(c, db) })
+}
+
+func addChannelTagHandler(c *gin.Context, db *sqlx.DB) {
+	var ct channelTag
+	if err := c.BindJSON(&ct); err != nil {
+		server_errors.LogAndSendServerError(c, err)
```
maxwedwards, line 24: Consider using the Wrap version of this to give a bit more context on what was just attempted

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	if ct.ChannelId != 0 {
+		if ct.NodeId == 0 {
+			server_errors.LogAndSendServerError(c, errors.New("Failed to find nodeId in the request."))
+			return
+		}
+		if ct.PeerId == 0 {
+			server_errors.LogAndSendServerError(c, errors.New("Failed to find peerId in the request."))
+			return
+		}
+	}
```
maxwedwards, line 48: These don't look like server _errors_ but look more like validation failures. There are a couple of other functions in the server_errrors package for bad requests and unprocessable entity. This looks more like an unprocessable entity to me.

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	if ct.PeerId != 0 {
+		corridor.FromPeerId = null.Int{NullInt64: sql.NullInt64{Int64: int64(ct.PeerId), Valid: true}}
+	}
+	if ct.ChannelId != 0 {
+		corridor.ChannelId = null.Int{NullInt64: sql.NullInt64{Int64: int64(ct.ChannelId), Valid: true}}
+	}
+	_, err := corridors.AddCorridor(db, corridor)
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
```
maxwedwards, line 62: Consider the Wrap version of this function call `WrapLogAndSendServerError`

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+		return
+	}
+	err = corridors.RefreshCorridorCacheByType(db, *corridors.GetCorridorTypeFromId(corridor.CorridorTypeId))
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
```
maxwedwards, line 67: Consider wrapping

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+		return
+	}
+	go func() {
+		err := GenerateChannelTag(db)
+		if err != nil {
+			log.Printf("Failed to generate channel tags. (%v)\n", err)
```
maxwedwards, line 73: Use zerolog for logging.

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
+		return
+	}
+	err = corridors.RefreshCorridorCacheByType(db, *corridors.GetCorridorTypeFromId(corridor.CorridorTypeId))
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
+		return
+	}
+	go func() {
```
maxwedwards, line 70: Is this sufficiently slow that we need to run it in it's own goroutine? The downside is that our users might not notice that there is a problem as they might not be scanning the logs.

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+
+func removeChannelTagHandler(c *gin.Context, db *sqlx.DB) {
+	channelTagId, err := strconv.Atoi(c.Param("channelTagId"))
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
+		return
+	}
+	ct, err := getChannelTag(db, channelTagId)
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
```
maxwedwards, line 87: Wrap?

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+		return
+	}
+	go func() {
+		err := GenerateChannelTag(db)
+		if err != nil {
+			log.Printf("Failed to generate channel tags. (%v)\n", err)
+		}
+	}()
```
maxwedwards, line 110: Same comments as Add, consider changing all LogAndSendServerError with WrapLogAndSendSeverError and consider not having the go routine if the channel tags can be generated in a reasonable time (hundreds of ms).

internal/channels/channel.go
```diff
@@ -102,26 +104,48 @@ type Channel struct {
...
 	CreatedOn         time.Time   `json:"createdOn" db:"created_on"`
 	UpdateOn          null.Time   `json:"updatedOn" db:"updated_on"`
 	LNDShortChannelID uint64      `json:"lndShortChannelId" db:"lnd_short_channel_id"`
 }
 
-func AddChannelRecordIfDoesntExist(db *sqlx.DB, channel Channel) error {
-	dbChannel, err := getChannel(db, channel.ShortChannelID)
+var ChannelIdByShortChannelId = make(map[string]int)
+
+func RefreshChannelsCache(db *sqlx.DB) error {
```
maxwedwards, line 117: I assume this is a WIP as it doesn't currently do anything

internal/channels/channel.go
```diff
@@ -102,26 +104,48 @@ type Channel struct {
...
 	LocalNodeId       int         `json:"localNodeId" db:"local_node_id"`
+	PeerId            int         `json:"peerId" db:"peer_id"`
 	CreatedOn         time.Time   `json:"createdOn" db:"created_on"`
 	UpdateOn          null.Time   `json:"updatedOn" db:"updated_on"`
 	LNDShortChannelID uint64      `json:"lndShortChannelId" db:"lnd_short_channel_id"`
 }
 
-func AddChannelRecordIfDoesntExist(db *sqlx.DB, channel Channel) error {
-	dbChannel, err := getChannel(db, channel.ShortChannelID)
+var ChannelIdByShortChannelId = make(map[string]int)
```
maxwedwards, line 115: No package level vars, especially exported ones. If data needs to be shared and can't be passed around then use go's channels (preferably) or Mutexes (preferably not). There are examples of both in the Torq codebase, happy to show you where.

internal/channels/database.go
```diff
@@ -3,37 +3,63 @@ package channels
...
+			return channels, nil
 		}
-		return nil, errors.Wrap(err, "Unable to execute SQL query")
+		return nil, errors.Wrap(err, database.SqlExecutionError)
+	}
+	newChannelIdByShortChannelId := make(map[string]int)
+	for _, c := range channels {
+		newChannelIdByShortChannelId[c.ShortChannelID] = c.ChannelDBID
+	}
+	ChannelIdByShortChannelId = newChannelIdByShortChannelId
```
maxwedwards, line 23: As above, no package level vars

cmd/torq/internal/torqsrv/server.go
```diff
@@ -2,39 +2,48 @@ package torqsrv
...
+	"unicode/utf8"
 )
 
 func Start(port int, apiPswd string, db *sqlx.DB, wsChan chan interface{}, restartLNDSub func() error) error {
 	r := gin.Default()
 
+	err := cache.ReinitializeCache(db)
+	if err != nil {
+		log.Fatal().Msg("Torq cannot be initialized (Loading caches in memory).")
+		return errors.Wrap(err, "Loading caches.")
```
maxwedwards, line 44: Either log or return error not both. If the error is returned it's up the caller to log or bubble it up further.

Here log.Fatal() will exit the whole application which could be right in this situation but perhaps leave it up the caller of Start() to decide what to do with the error.

internal/cache/cacheHelper.go
```diff
@@ -0,0 +1,17 @@
...
+)
+
+func ReinitializeCache(db *sqlx.DB) error {
+	log.Debug().Msg("Loading caches in memory.")
+	err := corridors.RefreshCorridorCache(db)
+	if err != nil {
+		return errors.Wrap(err, "Loading caches.")
+	}
+	return nil
+}
```
maxwedwards, line 17: Is this file needed at all? Move `refreshCorridorCache` to `Start()` function?

internal/channel_tags/database.go
```diff
@@ -0,0 +1,89 @@
...
+	tx, err := db.Begin()
+	if err != nil {
+		return errors.Wrap(err, database.SqlBeginTransactionError)
+	}
+
+	_, err = tx.Exec("DELETE FROM channel_tag WHERE tag_origin_id=$1;", corridor)
+	if err != nil {
+		if rb := tx.Rollback(); rb != nil {
+			log.Error().Msgf("Attempted to abort: %v (%v)", err, rb)
```
maxwedwards, line 64: Logging here is the right thing to do but you don't need to include err because that's going to get bubbled up on the next line.

When using zerolog, you can give err to the err() function rather than formatting it in using Msgf. See my error handling document for an example.
```suggestion
			log.Error().Err(err).Msg("Unable to rollback transaction")
```

internal/channel_tags/database.go
```diff
@@ -0,0 +1,89 @@
...
+	for _, channel := range channels {
+		for _, tag := range tags {
+			if corridors.GetBestCorridorFlag(corridors.CorridorKey{CorridorType: corridors.Tag, ReferenceId: tag.TagId, NodeId: channel.LocalNodeId, FromPeerId: channel.PeerId, ChannelId: channel.ChannelDBID}) == 1 {
+				_, err = tx.Exec(`
+					INSERT INTO channel_tag(node_id,peer_id,channel_db_id,tag_origin_id,tag_id,created_on)
+					VALUES ($1, $2, $3, $4, $5, $6);`,
+					channel.LocalNodeId, channel.PeerId, channel.ChannelDBID, corridor, tag.TagId, time.Now().UTC())
+				if err != nil {
+					if rb := tx.Rollback(); rb != nil {
+						log.Error().Msgf("Attempted to abort: %v (%v)", err, rb)
```
maxwedwards, line 77:
```suggestion
						log.Error().Err(rb).Msg("Unable to rollback transaction")
```

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+
+func RegisterChannelTagRoutes(r *gin.RouterGroup, db *sqlx.DB) {
+	r.POST("", func(c *gin.Context) { addChannelTagHandler(c, db) })
+	r.DELETE(":channelTagId", func(c *gin.Context) { removeChannelTagHandler(c, db) })
+}
+
+func addChannelTagHandler(c *gin.Context, db *sqlx.DB) {
+	var ct channelTag
+	if err := c.BindJSON(&ct); err != nil {
+		server_errors.SendBadRequestFromError(c, err)
```
maxwedwards, line 24: nit: could have wrapped error with a bit of context just incase the error out of BindJSON is not clear.
```suggestion
		server_errors.SendBadRequestFromError(c, errors.Wrap(err, "Parsing JSON"))
```

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	err = corridors.RefreshCorridorCacheByType(db, *corridors.GetCorridorTypeFromId(corridor.CorridorTypeId))
+	if err != nil {
+		server_errors.WrapLogAndSendServerError(c, err, "Refresh Corridor Cache By Type.")
+		return
+	}
+	go func() {
+		err := GenerateChannelTag(db)
+		if err != nil {
+			log.Printf("Failed to generate channel tags. (%v)\n", err)
```
maxwedwards, line 73:
```suggestion
			log.Error().Err(err).Msg("Failed to generate channel tags")
```

internal/channel_tags/routes.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	err = corridors.RefreshCorridorCacheByType(db, corridors.Tag)
+	if err != nil {
+		server_errors.WrapLogAndSendServerError(c, err, "Refresh Corridor Cache By Type.")
+		return
+	}
+	go func() {
+		err := GenerateChannelTag(db)
+		if err != nil {
+			log.Printf("Failed to generate channel tags. (%v)\n", err)
```
maxwedwards, line 108:
```suggestion
			log.Error().Err(err).Msg("Failed to generate channel tags")
```

internal/channels/channel.go
```diff
@@ -102,26 +103,33 @@ type Channel struct {
...
-	err = insertChannel(db, channel)
+	channel.PeerId = peerId
+	channelId, err := getChannelIdByShortChannelId(db, channel.ShortChannelID)
 	if err != nil {
 		return err
 	}
+	if channelId == 0 {
+		_, err = addChannel(db, channel)
+		if err != nil {
+			return err
```
maxwedwards, line 130: Maybe wrap?

internal/channels/handlers.go
```diff
@@ -72,8 +71,7 @@ func updateChannelsHandler(c *gin.Context, db *sqlx.DB) {
 	requestBody := updateChanRequestBody{}
 
 	if err := c.BindJSON(&requestBody); err != nil {
-		log.Error().Msgf("JSON binding the request body")
-		server_errors.WrapLogAndSendServerError(c, err, "JSON binding the request body")
+		server_errors.SendBadRequestFromError(c, err)
```
maxwedwards, line 74: Maybe consider wrapping error to give it a bit more context?

internal/channels/handlers.go
```diff
@@ -109,7 +107,7 @@ type BatchOpenResponse struct {
 func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
 	var batchOpnReq BatchOpenRequest
 	if err := c.BindJSON(&batchOpnReq); err != nil {
-		server_errors.WrapLogAndSendServerError(c, err, "JSON binding the request body")
+		server_errors.SendBadRequestFromError(c, err)
```
maxwedwards, line 110: maybe consider wrapping error to give it a bit more context?

internal/corridors/corridor.go
```diff
@@ -0,0 +1,253 @@
...
+)
+
+var MaximumPriority = getPriority(ToPeer)*2 - 1
+
+var Tag = CorridorType{0, "tag", 0}
+var AutoFee = CorridorType{1, "autoFee", 0}
+
+var CorridorTypes = []CorridorType{
+	Tag, AutoFee,
+}
```
maxwedwards, line 32: Two things:

I don't think these need to be exported as they are only used in this package.
No package level vars if we can avoid it. These look like constants.

maxwedwards (approved):
Looks good, happy for merge when tests are passing.

2Fast2BCn:
Don't merge ;-)
Henrik will fix the UI/UX a bit.
We also only support 2 tag colors so we better add some more

2Fast2BCn:
@ohenrik I rebased this branch onto main.
It's in a state where someone can pick it up once the design is finalized.

Issue #112, opened 2022-09-29
List Current Channel State (Simple channels table)


ohenrik:
lntop like live table listing all channels and have default filters for open/closed and pending channels. updating live similar to LN Top

PR #160, opened 2022-10-13
Feature/112 list current channels

kevenboo:
Related ticket -> https://github.com/lncapital/torq/issues/112

The in the screen bellow this is the page with the default columns and filter so only active channels will be displayed.

Screen Shot 2022-10-13 at 1 49 46 PM

I kept the reducers that were interacting with the columns and filters but if we want to display everything by default and not let the user add the columns I can remove them.

On the backend side the payload from LND is now returned in camelCase.


internal/channels/handlers.go
```diff
@@ -35,6 +40,32 @@ type pendingChannel struct {
 	PendingChannelPoint string `json:"pendingChannelPoint"`
 }
 
+type channelBody struct {
+	Active                *bool                 `json:"active"`
+	Gauge                 float64               `json:"gauge"`
+	RemotePubkey          *string               `json:"remotePubkey"`
+	ChannelPoint          *string               `json:"lndChannelPoint"`
```
maxwedwards, line 47: Sorry to be pedantic but can we call this LNDChannelPoint

internal/channels/handlers.go
```diff
@@ -35,6 +40,32 @@ type pendingChannel struct {
...
+	LNDShortChanID        *uint64               `json:"lndShortChannelId"`
+	ShortChanID           string                `json:"shortChannelID"`
```
maxwedwards, line 49: Can we call this ShortChannelId and in the json, shortChannelId with lower d?

internal/channels/handlers.go
```diff
@@ -35,6 +40,32 @@ type pendingChannel struct {
 	PendingChannelPoint string `json:"pendingChannelPoint"`
 }
 
+type channelBody struct {
+	Active                *bool                 `json:"active"`
+	Gauge                 float64               `json:"gauge"`
+	RemotePubkey          *string               `json:"remotePubkey"`
+	ChannelPoint          *string               `json:"lndChannelPoint"`
+	LNDShortChanID        *uint64               `json:"lndShortChannelId"`
```
maxwedwards, line 48: Nit: Can we lowercase the D on Id please?

internal/channels/routes.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"github.com/gin-gonic/gin"
+	"github.com/jmoiron/sqlx"
+)
+
+func RegisterChannelRoutes(r *gin.RouterGroup, db *sqlx.DB) {
+	r.POST("update", func(c *gin.Context) { updateChannelsHandler(c, db) })
+	r.POST("openbatch", func(c *gin.Context) { batchOpenHandler(c, db) })
+	r.GET("list", func(c *gin.Context) { getChannelListhandler(c, db) })
```
maxwedwards, line 11: Shouldn't this be a GET on `/api/channels/` rather than `/api/channels/list`. So just change to r.Get(""). The update is a bit suspicious why that's not just a POST on `/api/channels/` but not this PRs responsibility to fix that.

internal/channels/handlers.go
```diff
@@ -90,7 +121,79 @@ func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
...
 }
 
-func RegisterChannelRoutes(r *gin.RouterGroup, db *sqlx.DB) {
-	r.POST("update", func(c *gin.Context) { updateChannelsHandler(c, db) })
-	r.POST("openbatch", func(c *gin.Context) { batchOpenHandler(c, db) })
+func getChannelListhandler(c *gin.Context, db *sqlx.DB) {
+	var resp *lnrpc.ListChannelsResponse
+	var channelsBody []channelBody
+	nodes, _ := settings.GetActiveNodesConnectionDetails(db)
+	for _, node := range nodes {
```
maxwedwards, line 128: I've asked @ohenrik if this is ok to merge in now or if we should select a node based on nodeid that's passed in from the frontend.

internal/channels/handlers.go
```diff
@@ -90,7 +121,79 @@ func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
...
+		conn, err := lnd_connect.Connect(
+			node.GRPCAddress,
+			node.TLSFileBytes,
+			node.MacaroonFileBytes)
+		if err != nil {
+			log.Error().Err(err).Msgf("Failed to connect to lnd for node id: %v", node.LocalNodeId)
```
maxwedwards, line 134: The method below this line logs anyway so no need to log here, just delete.

internal/channels/handlers.go
```diff
@@ -90,7 +121,79 @@ func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
...
+			server_errors.WrapLogAndSendServerError(c, err, "Connect to node")
```
maxwedwards, line 135: Feel free to add the node id into the message if you want like you did in the above (unneeded) log call.

internal/channels/handlers.go
```diff
@@ -90,7 +121,79 @@ func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
...
+			return
+		}
+		defer conn.Close()
+		client := lnrpc.NewLightningClient(conn)
+		r, err := client.ListChannels(context.Background(), &lnrpc.ListChannelsRequest{})
+		if err != nil {
+			log.Error().Err(err).Msgf("Failed to connect to lnd for node id: %v", node.LocalNodeId)
```
maxwedwards, line 142: Again, no need to log as it's done below

internal/channels/handlers.go
```diff
@@ -90,7 +121,79 @@ func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
...
+		}
+
+		if resp == nil {
+			resp = r
+		} else {
+			resp.Channels = append(resp.Channels, r.Channels...)
+		}
+	}
+
+	for _, lndChannel := range resp.Channels {
```
maxwedwards, line 154: Personally, rather than putting all the channels into a *lnrpc.ListChannelsResponse, I'd just range over them and put them straight into a slice of channelBody. Saves having to append them all twice.

internal/channels/handlers.go
```diff
@@ -90,7 +121,79 @@ func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
...
+			CommitmentType:        &lndChannel.CommitmentType,
+			Lifetime:              &lndChannel.Lifetime,
+		}
+		channelsBody = append(channelsBody, chanBody)
+	}
+
+	c.JSON(http.StatusOK, channelsBody)
+}
+
+func sumHTLCs(htlcs *[]*lnrpc.HTLC) int64 {
```
maxwedwards, line 185: Do we need all these pointers? Can this just be a `[]lnrpc.HTLC`?

internal/channels/handlers.go
kevenboo: on it

internal/channels/handlers.go
kevenboo: on it

test/e2e/lnd/e2e_test.go
```diff
@@ -646,6 +647,14 @@ func TestPlaywrightVideo(t *testing.T) {
 
 	page.WaitForTimeout(1000)
 
+	page.Click("text=Channels")
+
+	page.WaitForTimeout(500)
+
+	page.Click("text=Default View")
+
+	page.WaitForTimeout(500)
```
maxwedwards, line 656: No generic time waits in tests please! I can see you have just copied what was before but this has to go.

web/src/Router.tsx
```diff
@@ -67,6 +68,8 @@ const authenticatedRoutes: RouteObject = {
         {
           path: routes.ANALYSE,
           children: [
+            { path: routes.CHANNELS, element: <ChannelsPage /> },
+            { path: routes.CHANNELS_CUSTOM_VIEW, element: <ChannelsPage /> },
```
maxwedwards, line 72: As we don't have custom views yet, I think we should drop the route for it

internal/channels/handlers.go
```diff
@@ -35,6 +40,32 @@ type pendingChannel struct {
...
+	PendingHtlcs          int64                 `json:"pendingHtlcs"`
+	TotalSatoshisSent     *int64                `json:"totalSatoshisSent"`
+	NumUpdates            *uint64               `json:"numUpdates"`
+	Initiator             *bool                 `json:"initiator"`
+	ChanStatusFlags       *string               `json:"chanStatusFlags"`
+	LocalChanReserveSat   *int64                `json:"localChanReserveSat"`
+	RemoteChanReserveSat  *int64                `json:"remoteChanReserveSat"`
+	CommitmentType        *lnrpc.CommitmentType `json:"commitmentType"`
+	Lifetime              *int64                `json:"lifetime"`
+	TotalSatoshisReceived *int64                `json:"totalSatoshisReceived"`
```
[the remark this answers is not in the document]
kevenboo: I though that instead of copying the data in new variables, i would only need to point to their values to save memory in case we have user with a lot of channels

web/src/apiSlice.ts
```diff
@@ -57,6 +57,13 @@ export const torqApi = createApi({
     getForwards: builder.query<any, GetForwardsQueryParams>({
       query: (params) => queryParamsBuilder("forwards", params, true),
     }),
+    getChannels: builder.query<any, void>({
```
maxwedwards, line 60: No anys please! There is a file called apiTypes or something where we have been putting types related to the API queries. If in doubt, ask @cryptosalomao as I think he has been refactoring in this area.

internal/channels/routes.go
kevenboo: Yes i can do that. I think when we did the call we talked about the endpoint being channels/list but yes i agree /channels is better

internal/channels/handlers.go
kevenboo: ok

web/src/constants/routes.ts
```diff
@@ -6,6 +6,8 @@ export const LOGOUT = "logout";
 export const ANALYSE = "analyse";
 export const FORWARDS = "forwards";
 export const FORWARDS_CUSTOM_VIEW = `${FORWARDS}/:viewId`;
+export const CHANNELS = "channels";
+export const CHANNELS_CUSTOM_VIEW = `${CHANNELS}/:viewId`;
```
maxwedwards, line 10: Again, I don't think we have custom views yet do we?

internal/channels/handlers.go
```diff
@@ -90,7 +121,79 @@ func batchOpenHandler(c *gin.Context, db *sqlx.DB) {
 	c.JSON(http.StatusOK, response)
 }
 
-func RegisterChannelRoutes(r *gin.RouterGroup, db *sqlx.DB) {
-	r.POST("update", func(c *gin.Context) { updateChannelsHandler(c, db) })
-	r.POST("openbatch", func(c *gin.Context) { batchOpenHandler(c, db) })
+func getChannelListhandler(c *gin.Context, db *sqlx.DB) {
+	var resp *lnrpc.ListChannelsResponse
+	var channelsBody []channelBody
+	nodes, _ := settings.GetActiveNodesConnectionDetails(db)
```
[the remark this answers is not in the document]
kevenboo: oh my bad I usually don't do that I implemented the error check after but I missed this one

internal/channels/handlers.go
kevenboo: on it

web/src/features/channels/ChannelsDataWrapper.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+  const sortBy = useAppSelector(selectSortBy);
+  const groupBy = useAppSelector(selectGroupBy) || "channels";
+  const filters = useAppSelector(selectFilters);
+
+  const [channels, columns] = useMemo(() => {
+    if (chanResponse.data?.length == 0) {
+      return [];
+    }
+    let channels = cloneDeep(chanResponse.data ? chanResponse.data : ([] as any[]));
```
maxwedwards, line 28: Does our clone function not do a deep clone? I'd rather not use lowdash if we can avoid it.

maxwedwards, line 28: No anys please

web/src/features/channels/ChannelsPage.tsx
```diff
@@ -0,0 +1,183 @@
+import { Link } from "react-router-dom";
+import {
+  Filter20Regular as FilterIcon,
+  ArrowSortDownLines20Regular as SortIcon,
+  ColumnTriple20Regular as ColumnsIcon,
+  ArrowJoin20Regular as GroupIcon,
+  Options20Regular as OptionsIcon,
+} from "@fluentui/react-icons";
+import Sidebar from "../sidebar/Sidebar";
```
maxwedwards, line 9: Same comment as before on relative paths

web/src/features/channels/ChannelsPage.tsx
```diff
@@ -0,0 +1,183 @@
...
+import { Clause, FilterCategoryType, FilterInterface } from "features/sidebar/sections/filter/filter";
+
+import TablePageTemplate, {
+  TableControlSection,
+  TableControlsButton,
+  TableControlsButtonGroup,
+  TableControlsTabsGroup,
+} from "../templates/tablePageTemplate/TablePageTemplate";
+import { useState } from "react";
+import { useAppDispatch, useAppSelector } from "../../store/hooks";
```
maxwedwards, line 20: Relative paths not needed

web/src/features/channels/ChannelsPage.tsx
```diff
@@ -0,0 +1,183 @@
...
+  updateSortBy,
+  selectGroupBy,
+  updateGroupBy,
+} from "./ChannelsSlice";
+import ColumnsSection from "../sidebar/sections/columns/ColumnsSection";
+import FilterSection from "../sidebar/sections/filter/FilterSection";
+import SortSection, { SortByOptionType } from "../sidebar/sections/sort/SortSectionOld";
+import GroupBySection from "../sidebar/sections/group/GroupBySection";
+import ChannelsDataWrapper from "./ChannelsDataWrapper";
+import { SectionContainer } from "../section/SectionContainer";
```
maxwedwards, line 37: Same comment on relative paths

web/src/features/channels/ChannelsPage.tsx
```diff
@@ -0,0 +1,183 @@
...
+    <TableControlSection>
+      <TableControlsButtonGroup>
+        <TableControlsTabsGroup>
+        </TableControlsTabsGroup>
+        <TableControlsButton onClickHandler={() => setSidebarExpanded(!sidebarExpanded)} icon={OptionsIcon} />
+      </TableControlsButtonGroup>
+    </TableControlSection>
+  );
+
+  const updateColumnsHandler = (columns: Array<any>) => {
```
maxwedwards, line 92: No Anys please

web/src/features/channels/ChannelsSlice.tsx
```diff
@@ -0,0 +1,195 @@
+import { createSlice, PayloadAction  } from "@reduxjs/toolkit";
+import { RootState } from "../../store/store";
+import { AndClause, FilterClause } from "../sidebar/sections/filter/filter";
+import { SortByOptionType } from "../sidebar/sections/sort/SortSectionOld";
+import { ColumnMetaData } from "../table/Table";
```
maxwedwards, line 5: No need for relative paths

maxwedwards (changes requested):
Great first PR! Covered a lot of the app!

Added comments on specific things, some of which I apologise you have copied from existing code but we now know better.

Any questions, please let me know

internal/channels/handlers.go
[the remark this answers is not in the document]
maxwedwards: I think especially if we do my suggestion below and just immediately copy the results into a slice of channelBody it will be fine. People don't have thousands of channels, hundreds maximum.

PR #161, opened 2022-10-13
connect peer & list peers

iulian-moraru:
connect peer & list peers endpoints
connect peer when opening channel

### DIFF
--- a/internal/channels/open_channel.go
+++ b/internal/channels/open_channel.go
@@ -13,7 +13,6 @@ import (
 	"github.com/lncapital/torq/pkg/lnd_connect"
 	"github.com/rs/zerolog/log"
 	"io"
-	"strings"
 )
 
 type OpenChannelRequest struct {
@@ -69,9 +68,9 @@ func OpenChannel(db *sqlx.DB, wChan chan interface{}, req OpenChannelRequest, re
 
 	ctx := context.Background()
 
-	//If host provided - check if peer and if needed connect peer
+	//If host provided - check if node is connected to peer and if not, connect peer
 	if req.NodePubKey != "" && req.Host != nil {
-		log.Debug().Msgf("Host provided. connect peer")
+		//log.Debug().Msgf("Host provided. connect peer")
 		if err := checkConnectPeer(client, ctx, req.NodeId, req.NodePubKey, *req.Host); err != nil {
 			return err
 		}
@@ -100,20 +99,13 @@ func OpenChannel(db *sqlx.DB, wChan chan interface{}, req OpenChannelRequest, re
 		}
 
 		if err != nil {
-			if strings.Contains(err.Error(), "is not online") {
-				log.Error().Msg("Peer is not online")
-				errNew := errors.New("Peer is not online. Provide full IP")
-				wChan <- errNew
-				return errNew
-			}
-			log.Error().Msgf("could not open channel: %v", err)
-			wChan <- errors.Newf("could not open channel: %v", err)
-			return err
+			log.Error().Msgf("Opening channel: %v", err)
+			return errors.Wrapf(err, "Opening channel")
 		}
 
 		r, err := processOpenResponse(resp)
 		if err != nil {
-			return err
+			return errors.Wrap(err, "Processing open response")
 		}
 		wChan <- r
 
@@ -122,16 +114,16 @@ func OpenChannel(db *sqlx.DB, wChan chan interface{}, req OpenChannelRequest, re
 
 func prepareOpenRequest(ocReq OpenChannelRequest) (r lnrpc.OpenChannelRequest, err error) {
 	if ocReq.NodeId == 0 {
-		return r, errors.New("Node id is missing")
+		return lnrpc.OpenChannelRequest{}, errors.New("Node id is missing")
 	}
 
 	if ocReq.SatPerVbyte != nil && ocReq.TargetConf != nil {
-		return r, errors.New("Cannot set both SatPerVbyte and TargetConf")
+		return lnrpc.OpenChannelRequest{}, errors.New("Cannot set both SatPerVbyte and TargetConf")
 	}
 
 	pubKeyHex, err := hex.DecodeString(ocReq.NodePubKey)
 	if err != nil {
-		return r, errors.New("error decoding public key hex")
+		return lnrpc.OpenChannelRequest{}, errors.New("error decoding public key hex")
 	}
 
 	//open channel request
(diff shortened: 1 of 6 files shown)


internal/channels/open_channel.go
```diff
@@ -94,6 +100,12 @@ func OpenChannel(db *sqlx.DB, wChan chan interface{}, req OpenChannelRequest, re
 		}
 
 		if err != nil {
+			if strings.Contains(err.Error(), "is not online") {
+				log.Error().Msg("Peer is not online")
+				errNew := errors.New("Peer is not online. Provide full IP")
+				wChan <- errNew
+				return errNew
+			}
```
maxwedwards, line 108: Rather than using `strings.Contains` can you see if you could use `errors.Is` or `errors.As`. Give me a shout if you want a hand with that. Also the rest of the error handling seems off to me. If you are logging + sending a custom error down the websocket, why are we also returning the error? What will the calling function do with that error? Something seems off there. Again, give me a shout if you want to talk it through.

internal/channels/open_channel.go
```diff
@@ -220,3 +232,34 @@ func translateChanPoint(cb []byte, oi uint32) (string, error) {
 
 	return fmt.Sprintf("%s:%d", ch.String(), oi), nil
 }
+
+func checkConnectPeer(client lnrpc.LightningClient, ctx context.Context, nodeId int, remotePubkey string, host string) (err error) {
+
+	peerList, err := peers.ListPeers(client, ctx, "/le")
+	if err != nil {
+		return err
```
maxwedwards, line 240: throw in an `return errors.Wrap(err, "Listing Peers")`?

internal/channels/open_channel.go
```diff
@@ -220,3 +232,34 @@ func translateChanPoint(cb []byte, oi uint32) (string, error) {
...
+		NodeId: nodeId,
+		LndAddress: peers.LndAddress{
+			PubKey: remotePubkey,
+			Host:   host,
+		},
+	}
+
+	_, err = peers.ConnectPeer(client, ctx, req)
+	if err != nil {
+		return err
```
maxwedwards, line 261: throw in an `return errors.Wrap(err, "Connect Peer")`?

internal/peers/connect_peer.go
```diff
@@ -0,0 +1,55 @@
...
+	"github.com/cockroachdb/errors"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"strings"
+)
+
+func ConnectPeer(client lnrpc.LightningClient, ctx context.Context, req ConnectPeerRequest) (r string, err error) {
+	connPeerReq, err := processRequest(req)
+
+	if err != nil {
+		return "", err
```
maxwedwards, line 14: error wrap with "process request"

internal/peers/connect_peer.go
```diff
@@ -0,0 +1,55 @@
...
+	}
+
+	_, err = client.ConnectPeer(ctx, &connPeerReq)
+	if err != nil {
+		if strings.Contains(err.Error(), "already connected") {
```
maxwedwards, line 19: Same as above, can you try to use `errors.Is` or `errors.As`

internal/peers/connect_peer.go
```diff
@@ -0,0 +1,55 @@
...
+	return "Peer connected", nil
+}
+
+func processRequest(req ConnectPeerRequest) (r lnrpc.ConnectPeerRequest, err error) {
+	if req.NodeId == 0 {
+		return lnrpc.ConnectPeerRequest{}, errors.Wrap(err, "Processing connect peer request")
+	}
+
+	if req.LndAddress.PubKey == "" || req.LndAddress.Host == "" {
+		return lnrpc.ConnectPeerRequest{}, errors.Wrap(err, "Missing LND Address or Host")
```
maxwedwards, line 34: again, err is nil so use `errors.New()` and explain what's wrong with having either pubKey or Host to be empty string

internal/peers/list_peers.go
```diff
@@ -0,0 +1,93 @@
...
+	LastFlapNs      int64              `json:"lastFlapNs"`
+	LastPingPayload []byte             `json:"lastPingPayload"`
+}
+
+func ListPeers(client lnrpc.LightningClient, ctx context.Context, latestErr string) (r []peer, err error) {
+
+	listPeerReq := lnrpc.ListPeersRequest{}
+	if latestErr != "/" {
+		listPeerReq.LatestError = true
+	}
```
maxwedwards, line 47: Can you talk me through what this is about?

internal/peers/handlers.go
```diff
@@ -0,0 +1,108 @@
...
+		return nil, errors.Wrap(err, "Connecting to LND")
+	}
+
+	return conn, nil
+
+}
+
+func RegisterPeersRoutes(r *gin.RouterGroup, db *sqlx.DB) {
+	r.POST("add", func(c *gin.Context) { connectPeerHandler(c, db) })
+	r.GET("list/:nodeId/*le", func(c *gin.Context) { listPeersHandler(c, db) })
```
maxwedwards, line 107: These routes aren't looking very RESTful. Can the add just be a POST to `/api/peers` and not `/api/peers/add` so just make this `r.POST("", func(....`.

And the list, shouldn't that be a URL like `/api/peers?localNodeId=1&le=blabla`?

maxwedwards (changes requested):
Very good, just a few comments. Get in touch if something doesn't make sense.

internal/channels/open_channel.go
iulian-moraru: list peers already returns errors.Wrap(err, "Listing Peers")

internal/channels/open_channel.go
iulian-moraru: ConnectPeer already returns errors.Wrap(err, "Connect Peer")

maxwedwards:
Error handling isn't quite right still. The errors.Is() call won't work like that with a string. It needs to have an error value, please verify for yourself that the errors.Is call is actually working as intended.

Have a look at this for examples of using errors.Is and errors.As: https://gosamples.dev/check-error-type/.

I've also published a guide to error handling in the Torq codebase: https://github.com/lncapital/torq/blob/main/docs/go_error_handling.md

As for your comments on not wrapping, I still think it would be best and you can differentiate them by for example having one being "Connect Peer" and the other being "LND: Connect peer" as the one is an internal function call and the other is to LND's api.

maxwedwards (changes requested):
Comments as above

PR #173, opened 2022-10-19
split channel route

### LINKED ISSUE #137: Split getChannelHistoryHandler into separate API endpoints
Split channel balance, event history and channel history into separate endpoints and request them individually in the frontend.

kevenboo:
Missing:
- Change payload to Camel
- Clanup comments
- Adapt frontend to new endpoints

Fixes #137


internal/channel_history/channel_balance.go
```diff
@@ -1,20 +1,21 @@
...
-	CapacityDiff     *int64    `db:"capacity_diff" json:"capacity_diff"`
+	InboundCapacity  int64     `db:"inbound_capacity" json:"inboundCapacity"`
+	OutboundCapacity int64     `db:"outbound_capacity" json:"outboundCapacity"`
+	CapacityDiff     *int64    `db:"capacity_diff" json:"capacityDiff"`
 }
 
 type ChannelBalance struct {
-	LNDShortChannelId string
-	Balances          []*Balance
+	LNDShortChannelId string     `json:"LndShortChannelId"`
```
2Fast2BCn, line 17: `json:"LndShortChannelId"` should be camelCase so `json:"lndShortChannelId"`

internal/channel_history/handlers.go
```diff
@@ -16,42 +17,52 @@ type ChannelHistory struct {
...
-	ChannelBalances []*ChannelBalance `json:"channel_balance"`
+	CountTotal *uint64 `json:"countTotal"`
 
 	// A list of channels included in this response
 	Channels []*channel               `json:"channels"`
 	History  []*ChannelHistoryRecords `json:"history"`
-	Events   []*ChannelEvent          `json:"events"`
+}
+
+func getChannelFrom(queryFrom string) (time.Time, error) {
```
2Fast2BCn, line 48: I cannot seem to find where these 3 methods are actually referenced.

internal/channel_history/handlers.go
```diff
@@ -91,23 +102,104 @@ func getChannelHistoryHandler(c *gin.Context, db *sqlx.DB) {
...
+
+type ChannelEventHistory struct {
+	Events []*ChannelEvent `json:"events"`
+}
+
+func getChannelEventHistoryHandler(c *gin.Context, db *sqlx.DB) {
+	var r ChannelEventHistory
+	from, err := time.Parse("2006-01-02", c.Query("from"))
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
```
2Fast2BCn, line 121: I think we decided to use the following for parameter handlers. (same for parameter to)
`server_errors.SendBadRequest(c, "Failed to find/parse ... in the request.")`

internal/channel_history/handlers.go
```diff
@@ -91,23 +102,104 @@ func getChannelHistoryHandler(c *gin.Context, db *sqlx.DB) {
...
+}
+
+type ChannelBalanceHistory struct {
+	// Channel balances over time
+	ChannelBalances []*ChannelBalance `json:"channelBalance"`
+}
+
+func getChannelBalanceHandler(c *gin.Context, db *sqlx.DB) {
+	var r ChannelBalanceHistory
+	from, err := time.Parse("2006-01-02", c.Query("from"))
```
2Fast2BCn, line 148: I think we decided to use the following for parameter handlers. (same for parameter to)
server_errors.SendBadRequest(c, "Failed to find/parse ... in the request.")

internal/channel_history/handlers.go
```diff
@@ -91,23 +102,104 @@ func getChannelHistoryHandler(c *gin.Context, db *sqlx.DB) {
...
 		server_errors.LogAndSendServerError(c, err)
 		return
 	}
 
+	c.JSON(http.StatusOK, r)
+}
+
+type ChannelBalanceHistory struct {
+	// Channel balances over time
+	ChannelBalances []*ChannelBalance `json:"channelBalance"`
```
2Fast2BCn, line 143: `json:"channelBalance"` should that be `json:"channelBalances"` (with an s) to keep it consistent?

internal/channel_history/handlers.go
```diff
@@ -91,23 +102,104 @@ func getChannelHistoryHandler(c *gin.Context, db *sqlx.DB) {
...
+		server_errors.LogAndSendServerError(c, err)
+		return
+	}
+
+	chanIds := strings.Split(c.Param("chanIds"), ",")
+
+	if chanIds[0] != "1" {
+
+		for _, chanId := range chanIds {
+			cb, err := getChannelBalance(db, string(chanId), from, to)
```
2Fast2BCn, line 164: `string(chanId)` it already is a string ?

internal/channel_history/handlers.go
```diff
@@ -91,23 +102,104 @@ func getChannelHistoryHandler(c *gin.Context, db *sqlx.DB) {
...
+	}
+	to, err := time.Parse("2006-01-02", c.Query("to"))
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
+		return
+	}
+
+	chanIds := strings.Split(c.Param("chanIds"), ",")
+
+	if chanIds[0] != "1" {
```
2Fast2BCn, line 161: shouldn't this be `if chanIds[0] != "" {`

internal/channel_history/handlers.go
```diff
@@ -91,23 +102,104 @@ func getChannelHistoryHandler(c *gin.Context, db *sqlx.DB) {
...
+	RebalancingCost *uint64 `json:"rebalancingCost"`
+	// Aggregated details about successful rebalancing (i.g. amount, cost, counts)
+	RebalancingDetails RebalancingDetails `json:"rebalancing"`
+}
+
+func getChannelReBalancingHandler(c *gin.Context, db *sqlx.DB) {
+	var r ChannelReBalancing
+	from, err := time.Parse("2006-01-02", c.Query("from"))
+	if err != nil {
+		server_errors.LogAndSendServerError(c, err)
```
2Fast2BCn, line 192: I think we decided to use the following for parameter handlers. (same for parameter to)
server_errors.SendBadRequest(c, "Failed to find/parse ... in the request.")

internal/channel_history/handlers.go
```diff
@@ -91,23 +102,104 @@ func getChannelHistoryHandler(c *gin.Context, db *sqlx.DB) {
...
+		}
+
+	}
+	c.JSON(http.StatusOK, r)
+}
+
+type ChannelReBalancing struct {
+	RebalancingCost *uint64 `json:"rebalancingCost"`
+	// Aggregated details about successful rebalancing (i.g. amount, cost, counts)
+	RebalancingDetails RebalancingDetails `json:"rebalancing"`
```
2Fast2BCn, line 185: Not sure if it's important but the name is inconsistent here `json:"rebalancing"` instead of `json:"rebalancingDetails"`

web/src/apiSlice.ts
```diff
@@ -53,11 +59,23 @@ export const torqApi = createApi({
   baseQuery: baseQueryWithRedirect,
   tagTypes: ["settings", "tableView", "localNodes", "channels"],
   endpoints: (builder) => ({
-    getFlow: builder.query<any, GetFlowQueryParams>({
-      query: (params) => queryParamsBuilder("flow", params), //
+    getFlow: builder.query<any, GetChannelHistoryQueryParams>({
```
2Fast2BCn, line 62: There is an `any` here not sure if it needs to be addressed now or when we refactor the `any`'s in bulk

2Fast2BCn (changes requested):
Looks very good I only saw some minor inconsistencies. I'm not sure they all need to be addressed but I just noticed the ones I could found.

internal/channel_history/handlers.go
maxwedwards: Agree, @kevenboo we have 2 new error types in server_errors. BadRequest (400) when you can't parse the request (like here) and Unprocessable Entity (422) when there is a validation failure or an Id is zero rather than a valid id for example. In the future hopefully we will respond differently on the front end based on these codes.

web/src/features/channel/ChannelPage.tsx
```diff
@@ -69,53 +86,64 @@ function ChannelPage(props: ChannelPageProps) {
...
+  const onChainCost = clone<ChannelOnchainCostResponse>(onChainCostResponse as ChannelOnchainCostResponse) || 0;
+
+  const { data: historyResponse } = useGetChannelHistoryQuery(queryParams);
+  const history = clone<ChannelHistoryResponse>(historyResponse as ChannelHistoryResponse) || null;
+
+  const { data: rebalancingResponse } = useGetChannelRebalancingQuery(queryParams);
+  const rebalancing = clone<ChannelRebalancingResponse>(rebalancingResponse as ChannelRebalancingResponse) || null;
+
+  const { data: eventResponse } = useGetChannelEventQuery(queryParams);
+  const event = clone<ChannelEventResponse>(eventResponse as ChannelEventResponse) || null;
```
maxwedwards, line 109: This doesn't look quite right to me, we shouldn't need to make clones of the data. Few extra notes also, we should prefer using `undefined` over `null` and there shouldn't be logic in the main component outside of a hook such as useEffect.

Can you talk me through why you are cloning the data rather than just using the response?

web/src/features/channel/DashboardPage.tsx
```diff
@@ -22,26 +34,31 @@ function ChannelPage() {
...
+
+  const { data, isLoading } = useGetFlowQuery(queryParams);
+  const { data: onChainCostResponse } = useGetChannelOnChainCostQuery(queryParams);
+  const onChainCost = clone<ChannelOnchainCostResponse>(onChainCostResponse as ChannelOnchainCostResponse) || 0;
+
+  const { data: historyQueryResponse } = useGetChannelHistoryQuery(queryParams);
+  const historyQuery = clone<ChannelHistoryResponse>(historyQueryResponse as ChannelHistoryResponse) || null;
+
+  const { data: rebalancingResponse } = useGetChannelRebalancingQuery(queryParams);
+  const rebalancing = clone<ChannelRebalancingResponse>(rebalancingResponse as ChannelRebalancingResponse) || null;
```
maxwedwards, line 53: Same comment as above on the clones.

web/src/utils/queryParamsBuilder.ts
```diff
@@ -12,9 +13,12 @@ const baseQueryDefaultPaginationParams: BaseQueryCollectionParams = {
...
 
 const parseEndpointDefinition = <T extends { [s: string]: unknown }>(
-  { endpoint, baseParam }: EndpointDefinition,
+  { endpoint, baseParam, suffixEndpoint }: EndpointDefinition,
   params: T
-): string => `${endpoint}/${params[baseParam]}`;
+): string => {
+ if (!suffixEndpoint) return `${endpoint}/${params[baseParam]}`
+ return  `${endpoint}/${params[baseParam]}/${suffixEndpoint}`
+};
```
maxwedwards, line 21: I'm not a fan of this, what do we do if we go another level deep? `suffixSuffixEndpoint`? I see that you are operating inside the constraints of the existing function but I think this function needs to change to something more flexible. I wouldn't hold up this PR for this and will discuss it as part of the team.

maxwedwards (commented):
Nice PR! Just a few extra comments in addition to @2Fast2BCn. Happy to talk about any of it on slack or call.

internal/channel_history/channel_balance.go
kevenboo: Fixed

Issue #182, opened 2022-10-24
Saving views for all table pages (channels, payments, etc.)


ohenrik:
Example from the forwards page:
![Screenshot 2022-10-24 at 13 49 36](https://user-images.githubusercontent.com/647617/197519044-614e74de-afb1-4840-a76f-ae03a8720dc7.png)

The backend for saved views needs an added column to separate out the different pages of the view.

PR #190, opened 2022-10-31
Feaure/182 save channels view

kevenboo:
Related ticket -> https://github.com/lncapital/torq/issues/182
- Transaction pages have now  the possibility to save and update views
- There were an issue when switching from pages with views when the result are cached.
Like going from "forwards" to "transactions" or in transactions from "onchain" to "invoices"

Screen Shot 2022-10-31 at 11 45 52 AM
All views are set to be the last page / tab visited in this case "payments"

This will cause the page to load the wrong view information.

The fix applied was to load the views from the DB and store them.
We might have a better way to clear the cache or not save a specific view for all pages.


web/src/features/viewManagement/ViewsPopover.tsx
```diff
@@ -171,7 +188,7 @@ function ViewsPopover() {
     updateTableViewsOrder(order);
   };
 
-  const button = <TabButton title={views[selectedView].title} dropDown={true} />;
+  const button = <TabButton title={views[selectedView]?.title} dropDown={true} />;
```
ohenrik, line 191: I think this can potentially fail, since TabButton requires a title and `views[selectedView]?.title` can potentially be undefined. I'm not sure how/why Typescript allows this.

database/migrations/000044_add_column_page_to_table_view.up.psql
```diff
@@ -0,0 +1 @@
+ALTER TABLE table_view ADD COLUMN IF NOT EXISTS page VARCHAR(64);
```
maxwedwards, line 1: Is there a reason why it's 64? Can we just use the TEXT datatype?

ohenrik (commented):
Great PR, just some potential issues

database/migrations/000045_update_existing_table_view.up.psql
```diff
@@ -0,0 +1 @@
+UPDATE table_view SET page = 'forwards' WHERE page IS NULL;
```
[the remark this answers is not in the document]
kevenboo: Fixed. Merged both in the same migration :)

web/src/features/viewManagement/ViewsPopover.tsx
maxwedwards: After investigation, I believe we can remove the `?` here

PR #192, opened 2022-10-31
refactor queryparamsbuilder

maxwedwards:


maxwedwards:
Sorry about the formatting changes, someone doesn't have prettier installed and formatting!

web/src/apiSlice.ts
```diff
@@ -61,26 +62,26 @@ export const torqApi = createApi({
...
   endpoints: (builder) => ({
-    getFlow: builder.query<FlowData[], GetChannelHistoryQueryParams>({
-      query: (params) => queryParamsBuilder("flow", params),
+    getFlow: builder.query<FlowData[], GetFlowQueryParams>({
+      query: (params) => "flow" + queryParamsBuilder(params),
     }),
-    getChannelHistory: builder.query<ChannelHistoryResponse, GetChannelHistoryQueryParams>({
-      query: (params) => queryParamsBuilder({ endpoint: "channels", baseParam: "chanIds", suffixEndpoint: "history" }, params),
+    getChannelHistory: builder.query<ChannelHistoryResponse, GetChannelHistoryData>({
+      query: (data) => `channels/${data.params.chanId}/history` + queryParamsBuilder(data.queryParams),
```
ohenrik, line 69: This :

     `channels/${data.params.chanId}/history` + queryParamsBuilder(data.queryParams)

Can be written like this:

     `channels/${data.params.chanId}/history&{queryParamsBuilder(data.queryParams)}`

But I'm not sure if there is any benefits to or if it's cleaner, but since we are using that type of string interpolation on that line I think we should stick to using it (i.e. the second version). or switch to always using the pluss sign.

web/src/features/channel/ChannelPage.tsx
```diff
@@ -372,16 +369,18 @@ function ChannelPage(props: ChannelPageProps) {
...
+              <div className={classNames(styles.rowValue)}>{d3.format(".1%")(historyCountInOut)}</div>
             </div>
           </div>
         </div>
 
         <div className={classNames(styles.pageRow, styles.eventSummary)}>
           <div className={styles.shortColumn}>
-            <EventsCard events={event as ChannelEventResponse} selectedEvents={selectedEvents} channels={history as ChannelHistoryResponse} />
+            <EventsCard
+              events={event as ChannelEventResponse}
```
ohenrik, line 380: Are these uses `as` here a quick fix or is this how we are supposed to make this work? It seems like `useGetChannelEventQuery ` should have a typed return of `ChannelHistoryResponse `

ohenrik (commented):
Looks good except for some questions about potential issues/improvements

web/src/features/channel/ChannelPage.tsx
maxwedwards: Pretty sure I'm only making a formatting change here.

web/src/apiSlice.ts
maxwedwards: I don't have a strong opinion but I did intentionally do it that way to highlight the difference between params in the URL and query string params. Happy to change if you prefer it all in the string template.

web/src/features/channel/ChannelPage.tsx
ohenrik: That's correct :)

web/src/apiSlice.ts
ohenrik: I see, that makes sense I guess :)

web/src/utils/queryParamsBuilder.ts
```diff
@@ -1,54 +1,17 @@
...
+    if (params[key] === undefined) continue;
+    if (typeof params[key] === "object") {
+      paramObject[key] = JSON.stringify(params[key]);
+      continue;
+    }
+    paramObject[key] = String(params[key]);
+  }
+
+  if (isPaginable) {
+    paramObject = { limit: "100", offset: "0", ...paramObject };
```
kevenboo, line 13: Maybe not now but we could put those values in a constant file or even as env variable. What do you think ?

web/src/features/channel/ChannelPage.tsx
kevenboo: For this part, **getChannelEvent** was set to return a type **ChannelEventResponse** but when calling the function useGetChannelEventQuery,  "event" is set as any that why the type is set here.

kevenboo (approved):
The PR looks good to me.

PR #193, opened 2022-10-31
Lint/staticcheck

maxwedwards:


maxwedwards:
@ohenrik if you checkout this PR you will see that there are two checks to do with LNDs DestAddresses being deprecated. Please can you decide what needs to be done on those and fix this PR so that the tests work and then we can merge it.

ohenrik:
> @ohenrik if you checkout this PR you will see that there are two checks to do with LNDs DestAddresses being deprecated. Please can you decide what needs to be done on those and fix this PR so that the tests work and then we can merge it.

That's neat! I did not know we are using a deprecated field here. We should instead iterate through the `OutputDetails`   (`[]*OutputDetail`) and get the addresses from there. I'll create an issue for it.

pkg/lnd/transactions.go
```diff
@@ -137,24 +148,14 @@ func storeTransaction(db *sqlx.DB, tx *lnrpc.Transaction) error {
...
-			tx.TxHash,
-			tx.Amount,
-			tx.NumConfirmations,
-			tx.BlockHash,
-			tx.BlockHeight,
-			tx.TotalFees,
-			pq.Array(tx.DestAddresses),
-			tx.RawTxHex,
-			tx.Label)
+		return errors.Wrapf(err, `inserting transaction`)
```
maxwedwards, line 157: nice

ohenrik: 🎉

Issue #144, opened 2022-10-05
Incorrect password trigges a complete page refresh and missing toast


ohenrik:
If the user enters the incorrect password, the entire pag reloads. There also isn't any message letting the user know if the password was incorrect or if the the frontend could not reach the backend.

PR #195, opened 2022-11-01
Send error when wrong password

kevenboo:
Fixes https://github.com/lncapital/torq/issues/144

When password is wrong
Screen Shot 2022-11-01 at 11 29 56 AM

When api not reachable
Screen Shot 2022-11-01 at 11 29 32 AM

### DIFF
--- a/web/src/apiSlice.ts
+++ b/web/src/apiSlice.ts
@@ -24,6 +24,7 @@ import type {
   GetTableViewQueryParams,
   GetFlowQueryParams,
   GetChannelHistoryData,
+  LoginResponse,
 } from "types/api";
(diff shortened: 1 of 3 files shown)


ohenrik:
Tip: If you create a PR with "Fixes #<issue number>"  instead of "Related ticket", then Github automatically closes the issue when we merge the PR.

Screenshot 2022-11-01 at 12 47 55

kevenboo:
> Tip: If you create a PR with "Fixes #" instead of "Related ticket", then Github automatically closes the issue when we merge the PR.
> 
> Screenshot 2022-11-01 at 12 47 55
> 
> Screenshot 2022-11-01 at 12 49 46

Nice thank you. I will do that in the Future

Issue #171, opened 2022-10-18
Feature Request: Support ping systems like Amboss.


2Fast2BCn:
Support ping systems like Amboss.

2Fast2BCn:
TODO: Setup in UI.

PR #219, opened 2022-11-15
#171: Support ping systems like Amboss.

2Fast2BCn:


cmd/torq/internal/amboss_ping/server.go
```diff
@@ -0,0 +1,55 @@
...
+	"time"
+
+	"github.com/cockroachdb/errors"
+	"github.com/lightningnetwork/lnd/lnrpc"
+	"github.com/rs/zerolog/log"
+
+	"google.golang.org/grpc"
+)
+
+const ambossUrl = "https://api.amboss.space/graphql"
```
ohenrik, line 19: Should this perhaps be moved inside the function between line 45 and 56? So we avoid a global variable?

internal/settings/nodeConnectionDetails.go
```diff
@@ -82,6 +83,7 @@ func AddNodeToDB(db *sqlx.DB, implementation commons.Implementation,
 		TLSDataBytes:      tlsDataBytes,
 		MacaroonDataBytes: macaroonDataBytes,
 		CreateOn:          time.Now().UTC(),
+		PingSystem:        0,
```
ohenrik, line 86: What does PingSystem 0 mean? Should this perhaps be an enumerator?

PR #233, opened 2022-11-22
use database for static channel data

### LINKED ISSUE #231: Channels page Improvment

kevenboo:
Fixes #231 
- Use internal db info for the channel routing policy instead of calling LND
- update page layout
Screenshot 2022-11-23 at 9 43 55 AM


internal/channels/handlers.go
```diff
@@ -225,18 +233,18 @@ func getChannelListhandler(c *gin.Context, db *sqlx.DB) {
...
 				NumUpdates:                   channel.NumUpdates,
 				Initiator:                    channel.Initiator,
 				ChanStatusFlags:              channel.ChanStatusFlags,
 				CommitmentType:               channel.CommitmentType,
 				Lifetime:                     channel.Lifetime,
 				MempoolSpace:                 MEMPOOL + stringLNDShortChannelId,
-				AmbossSpace:                  AMBOSS + shortChannelId,
+				AmbossSpace:                  AMBOSS + channelRoutingPolicy[0].ShortChannelId,
 				OneMl:                        ONEML + stringLNDShortChannelId,
 			}
```
ohenrik, line 250: It seems like github won't let me comment on code that has not changed, but we need to switch out the node alias as well (get that from the node_event table).

ohenrik: Screenshot 2022-11-22 at 10 14 47

internal/channels/database.go
```diff
@@ -23,6 +23,25 @@ func GetAllChannels(db *sqlx.DB) (channels []Channel, err error) {
...
+func GetRoutingPolicy(chanId uint64, db *sqlx.DB) ([]ChannelPolicy, error) {
+	cp := []ChannelPolicy{}
+	err := db.Select(&cp, `
+	SELECT time_lock_delta, min_htlc, max_htlc_msat, fee_base_msat, fee_rate_mill_msat, short_channel_id
+	FROM routing_policy rp
+	LEFT JOIN channel c
+	ON rp.channel_id = c.channel_id
+	WHERE c.lnd_short_channel_id=$1
+	ORDER BY ts DESC
+	LIMIT 1;`, chanId)
```
ohenrik, line 35: You are here just fetching the first policy, the problem with that is that there is one for each direction. So you should have two queries here. one for the inbound policy (the remote nodes fees to you) and one for the outbound policy (fees from you to your peer).

When the announcing_node_id = our node id then it is an outbound policy. When connecting_node_id = our node id, then it is the inbound policy.

internal/channels/handlers.go
```diff
@@ -225,18 +233,18 @@ func getChannelListhandler(c *gin.Context, db *sqlx.DB) {
...
-				BaseFeeMsat:                  channelFee.Node1Policy.FeeBaseMsat,
-				MinHtlc:                      channelFee.Node1Policy.MinHtlc,
-				MaxHtlcMsat:                  channelFee.Node1Policy.MaxHtlcMsat,
-				TimeLockDelta:                channelFee.Node1Policy.TimeLockDelta,
-				FeeRatePpm:                   channelFee.Node1Policy.FeeRateMilliMsat,
+				BaseFeeMsat:                  channelRoutingPolicy[0].BeeBaseMsat,
+				MinHtlc:                      channelRoutingPolicy[0].MinHtlc,
+				MaxHtlcMsat:                  channelRoutingPolicy[0].MaxHtlcMsat,
+				TimeLockDelta:                channelRoutingPolicy[0].TimeLockDelta,
+				FeeRatePpm:                   channelRoutingPolicy[0].FeeRateMillMsat,
```
ohenrik, line 240: As per the comment above, here you should list both the inbound and the outbound policies (so twice the amount of columns)

ohenrik (changes requested):
Some changes needed, but looks good :) I had not cleared up a misunderstanding from the previous implementation.

ohenrik (approved):

ohenrik:
Looks great! ⚡ Merging

PR #256, opened 2022-12-13
Check latest for node

maxwedwards:
Checking latest imported data did not take into account which node it was for


2Fast2BCn (approved):
Very nice find!

PR #263, opened 2022-12-14
Network select

maxwedwards:


maxwedwards:
Other endpoints and possibly database queries need to be moved over to use this network select. Working for Forwards only at present but it's an improvement so valid to go in.

Issue #268, opened 2022-12-14
Channel Management - policy update of one feature triggers other updates


feelancer21:
Version 0.15.2

You have to set the basefee, feerate and timelockdelta if you update a channel policy. Otherwise defaults of 0sat, 0ppm and 18blocks are used.

1) I updated the feerate from 300ppm to 250ppm via UI. But also the TimeLockDelta and BaseFee changed

Old policy
```
  "node1_policy": {
        "time_lock_delta": 99,
        "min_htlc": "9999999",
        "fee_base_msat": "9999",
        "fee_rate_milli_msat": "999",
        "disabled": false,
        "max_htlc_msat": "99999999",
        "last_update": 1671050726

```

New policy
```
    "node1_policy": {
        "time_lock_delta": 18,
        "min_htlc": "9999999",
        "fee_base_msat": "0",
        "fee_rate_milli_msat": "250",
        "disabled": false,
        "max_htlc_msat": "99999999",
        "last_update": 1671050807
    },

```

2) I updated the TimeLockDelta and it sets FeeRate and BaseFee to zero

old policy
```
    "node1_policy": {
        "time_lock_delta": 99,
        "min_htlc": "9999999",
        "fee_base_msat": "9999",
        "fee_rate_milli_msat": "999",
        "disabled": false,
        "max_htlc_msat": "99999999",
        "last_update": 1671050835
    },
```

new policy
```
    "node1_policy": {
        "time_lock_delta": 144,
        "min_htlc": "9999999",
        "fee_base_msat": "0",
        "fee_rate_milli_msat": "0",
        "disabled": false,
        "max_htlc_msat": "99999999",
        "last_update": 1671050865
    },
```

2Fast2BCn:
Thx for reporting!

PR #270, opened 2022-12-15
#268: policy update of one feature triggers other updates.

2Fast2BCn:


internal/channels/update_channel.go
```diff
@@ -76,6 +69,15 @@ func createPolicyRequest(req commons.UpdateChannelRequest) (r *lnrpc.PolicyUpdat
 		updChanReq.Scope = &lnrpc.PolicyUpdateRequest_Global{Global: true}
 	}
 
+	//Minimum supported value for TimeLockDelta is 18
+	if req.TimeLockDelta != nil {
```
ohenrik, line 73: Did you test this manually? I thought LND required that this is not nil.

ohenrik (changes requested):

internal/channels/update_channel.go
ohenrik: It might be better to set the previous timelockDelta as the default. This can be done in the frontend too if that is easier.

2Fast2BCn: OK, I wasn't aware of this LND api requirement.

2Fast2BCn: I tested this time ;-)

PR #278, opened 2022-12-21
Network select

maxwedwards:


internal/channel_history/rebalancing_history.go
```diff
@@ -38,8 +38,9 @@ func getRebalancingCost(db *sqlx.DB, nodeIds []int, from time.Time, to time.Time
 			WHERE status = 'SUCCEEDED' AND
 				htlcs->-1->'route'->'hops'->-1->>'pub_key' = ANY($1) AND
```
ohenrik, line 39: I know this is not part of your change, but I think there is a small problem with this query caused by the fact that we now have multiple nodes. 

The problem is that we check if the last hop in the payments is to _one of_ our own nodes. But a rebalance of a channel is a payment from your own node to the same node, not any of the other nodes you own.

This might not be that big of a problem, and potentially not a problem at all. Since if you send money between your nodes directly there is no cost. If you are using a non direct route, then you would probably like to include it as a rebalance cost anyway. Since it is... 

So I would say that we do not change this now. Maybe just add a comment about this (for example just copy fasta this text)

internal/channels/database.go
```diff
@@ -24,18 +24,18 @@ func GetAllChannels(db *sqlx.DB) (channels []Channel, err error) {
...
-		return nil, errors.Wrapf(err, "Running getChannels query all: %v, channelIds: %v", all, channelIds)
+		return nil, errors.Wrapf(err, "Running getChannels query")
 	}
 	var r []*Channel
 	for rows.Next() {
 		c := &Channel{}
 		err = rows.StructScan(&c)
 		if err != nil {
-			return nil, errors.Wrapf(err, "Running getChannels query StructScan all: %v, channelIds: %v", all, channelIds)
+			return nil, errors.Wrapf(err, "Running getChannels query StructScan")
```
ohenrik, line 38: Would this data not be useful to have? Knowing which channel ids where requested is pretty useful as the frontend might have requested something silly and the amount of data is not large and not sensitive.

ohenrik (approved):
Looks good to me except for two questions.

internal/channels/database.go
maxwedwards: I can revert, just in general not a fan of runtime data being in logs

internal/channel_history/rebalancing_history.go
2Fast2BCn: table `payment` has `incoming_channel_id`, `outgoing_channel_id` and `rebalance_amount_msat`.
this data is populated by the provided htlc route.

Reading your comment makes me realize `rebalance_amount_msat` might be also suffering from this problem.
`rebalance_amount_msat` is populated when `incoming_channel_id` is not null.
However `incoming_channel_id` might be from a secondary node.

We should however have all the needed data to fix this with an update statement. Once we fixed this we should be able to simplify this query.

PR #308, opened 2023-01-11
Fix timestamp issue in channel history query

2Fast2BCn:


ohenrik:
what bug did this cause?

2Fast2BCn:
This should fix the Forwards Summary page.

PR #334, opened 2023-01-19
Add ability to use alternative to vector.

2Fast2BCn:


2Fast2BCn:
With this code we can spin up vector in dev mode and debug both sides.

PR #356, opened 2023-01-25
Workflow visibility settings

maxwedwards:


web/src/pages/WorkflowPage/workflowApi.ts
```diff
@@ -60,6 +61,13 @@ export const workflowApi = torqApi.injectEndpoints({
       }),
       invalidatesTags: ["workflow"],
     }),
+    updateNodeVisibilitySettings: builder.mutation<void, UpdateWorkflowNodeVisibilitySettingsRequest>({
+      query: (body: UpdateWorkflowNodeVisibilitySettingsRequest) => ({
+        url: `workflows/nodes/${body.workflowVersionNodeId}/visibility-settings`,
+        method: "PUT",
+        body: body.visibilitySettings,
+      }),
+    }),
```
ohenrik, line 70: I was first thinking you where missing `invalidatesTags: ["workflow"],` here, but I think it's not needed, since the position is updated in the frontend directly.

maxwedwards: That was my thinking too. It doesn't need to force a refetch.

PR #371, opened 2023-01-29
Some wrong labels.

2Fast2BCn:


ohenrik:
We need to migrate the saved views to make this work without breaking everyone's views. So we need to wait with merging this

ohenrik (changes requested):
We need to migrate (and perhaps change the data model) for existing views.

PR #378, opened 2023-02-01
tableView v3

2Fast2BCn:


ohenrik (commented):
I think we can simplify this a lot by using the original endpoints, let's talk through the changes

ohenrik:
Ready to merge if the update view (save view) is now again updating and not creating a new view.

PR #396, opened 2023-02-05
Fix help for --db.password

mmilata:
Also fix formatting of some other options to be consistent with the rest.


ohenrik:
Thank you @mmilata!

PR #398, opened 2023-02-05
Allow connecting to database through unix socket

mmilata:
Currently it is not possible to use `--db.host=/run/postgresql` because the slashes do not work with the URI-like connection string. This PR switches to alternative format. Using unix sockets has the advantage that it works with [peer authentication ](https://www.postgresql.org/docs/current/auth-peer.html) method, which makes it possible not to use database password at all.


2Fast2BCn:
Thx for the PR.
I hope you are ok with the fact I incorporated your changes into https://github.com/lncapital/torq/pull/402

mmilata:
Perfect, thanks!

PR #411, opened 2023-02-07
Improved validation messages from server errors

maxwedwards:


maxwedwards:
This is a WIP but also please feel free to merge so that other changes can start to benefit from the new structure. I think I will add arguments to the errror messages as per Kobe's suggestion.

PR #435, opened 2023-02-13
NPM Cleanup

maxwedwards:


maxwedwards:
I'm hopeful that this will correct the dependabot alert about json5.

It does remove a really old (6 year) package that types css modules but that has never worked on for me anyway and I don't think it's a big loss.

laitang2000 (approved):

PR #447, opened 2023-02-17
Workflow node for rebalancing

2Fast2BCn:


ohenrik (changes requested):
Before we merge this in we need to think about existing workflows. We need to:
1. Migrate all workflows to use a data source (or remove link to the triggers).
2. Disable all workflows
3. Add version information to Intercom so that we can show an alert to users who upgrades that they need to check their workflows.
4. Push all workflow nodes 400px to the right (to simplify adding data source)

These are known changes, so I'm just adding this list here so we don't forget.

internal/channels/update_channel.go
```diff
@@ -50,3 +50,45 @@ func SetRoutingPolicyWithTimeout(request commons.RoutingPolicyUpdateRequest,
...
+				Status:  commons.Inactive,
+				Message: message,
+				Error:   message,
+			},
+		}
+	}
+	responseChannel := make(chan commons.RebalanceResponse, 1)
+	request.ResponseChannel = responseChannel
+	rebalanceRequestChannel <- request
+	time.AfterFunc(commons.LIGHTNING_COMMUNICATION_TIMEOUT_SECONDS*time.Second, func() {
```
maxwedwards, line 82: This doesn't look safe to me. Normally we would use a context for cancelling something. This doesn't appear to cancel the original job but instead tries to get in there first and write to the response channel. The original job is still going to send something down the channel and what happens then?

internal/automation/managedRebalance.go
```diff
@@ -291,6 +326,12 @@ func isValidRequest(managedRebalance ManagedRebalance) bool {
 		return false
 	}
 	switch managedRebalance.Type {
+	case WRITE_REBALANCER:
+		fallthrough
```
maxwedwards, line 330: No need for fallthrough, it will do unnecessary work.

internal/automation/managedRebalance.go
```diff
@@ -43,8 +43,8 @@ type ManagedRebalance struct {
 }
 
 func ManagedRebalanceCache(ch chan ManagedRebalance, ctx context.Context) {
-	// rebalancers = map["workflow"]map[workflowVersionNodeId]Rebalancer
-	rebalancers := make(map[commons.RebalanceRequestOrigin]map[int]*Rebalancer)
+	// rebalancers = map["workflow"]map[workflowVersionNodeId][focusChannelId]Rebalancer
+	rebalancers := make(map[commons.RebalanceRequestOrigin]map[int]map[int]*Rebalancer)
```
maxwedwards, line 47: Reading through some code I think it might be useful to use specific types even if they are only aliases for int as the key on maps to help the reader know what it's a map of. Especially like in this instance where it's a map of map.

internal/automation/managedRebalance.go
```diff
@@ -211,19 +206,17 @@ func processResultsCache(
...
-							(managedRebalance.Status == nil || *managedRebalance.Status == rebalanceResults[i].Status) {
-							if latest.RebalanceId == 0 {
-								latest = rebalanceResults[i]
-								break
-							}
-							if latest.UpdateOn.Before(rebalanceResults[i].UpdateOn) {
-								latest = rebalanceResults[i]
-								break
-							}
+				for i := len(rebalanceResults) - 1; i >= 0; i-- {
```
maxwedwards, line 209: Probably worth explaining why you have to iterate in reverse

internal/automation/managedRebalance.go
```diff
@@ -43,8 +43,8 @@ type ManagedRebalance struct {
 }
 
 func ManagedRebalanceCache(ch chan ManagedRebalance, ctx context.Context) {
```
maxwedwards, line 45: Why do we need a cache? Would it be possible to do single threaded or using postgres?

maxwedwards:
I know this appears to have been merged in already but we should come up with an alternative way to trigger rebalances rather than add more things to the pub sub system.

maxwedwards:
Again already merged in but `commons.REBALANCE_MAXIMUM_CONCURRENCY`,	`commons.REBALANCE_RUNNER_TIMEOUT_SECONDS`, `commons.REBALANCE_ROUTES_TIMEOUT_SECONDS`, `commons.REBALANCE_PAY_TIMEOUT_SECONDS` do these things need to be in commons? They seem like they only apply here.

Ideally commons wouldn't exist at all, if it has to exist then it should have the minimum amount of things in it.

internal/channels/update_channel.go
2Fast2BCn: Removed all buffered channels and replaced this section with a loop in PR: https://github.com/lncapital/torq/pull/470

2Fast2BCn:
> Again already merged in but `commons.REBALANCE_MAXIMUM_CONCURRENCY`, `commons.REBALANCE_RUNNER_TIMEOUT_SECONDS`, `commons.REBALANCE_ROUTES_TIMEOUT_SECONDS`, `commons.REBALANCE_PAY_TIMEOUT_SECONDS`, `commons.REBALANCE_QUEUE_TICKER_SECONDS ` do these things need to be in commons? They seem like they only apply here.
> 
> Ideally commons wouldn't exist at all, if it has to exist then it should have the minimum amount of things in it.

Must be from my Java background. I'm used to centralizing all static parameters into a "Class" so you have a good overview without having to look for them in different packages of the source tree. If the code convention in golang is different then I'll relocate those into their own packages.

internal/automation/managedRebalance.go
2Fast2BCn: That is a very good idea!

internal/automation/managedRebalance.go
2Fast2BCn: Added a comment.

internal/automation/managedRebalance.go
2Fast2BCn: I don't understand what you mean here.
It will check if `IncomingChannelId` and `OutgoingChannelId` are both 0?

maxwedwards: What I mean is that fallthrough means it will check the following cases. But if the case WRITE_REBALANCER has matched then the others wont match so why are we checking them?

PR #459, opened 2023-02-21
introduce debug level

ziggie1984:
- backend main: adding loglevels
- backend workflow: adding more log messages
- lnd: fix small log bug and change log message


maxwedwards:
Thanks, nice improvement

PR #461, opened 2023-02-21
frontend: use localstorage for global network selection

ziggie1984:


2Fast2BCn:
I haven't looked into the detail of this PR but the services endpoint returns the networks Torq has configuration for, maybe we can incorporate this?

web/src/features/navigation/NetworkSelector.tsx
```diff
@@ -30,6 +31,7 @@ function NetworkSelector() {
             buttonColor={activeNetwork === Network.MainNet ? ColorVariant.success : ColorVariant.primary}
             onClick={() => {
               dispatch(setActiveNetwork(Network.MainNet));
+              setItem("activeNetwork", Network.MainNet);
```
ohenrik, line 34: By setting the local storage here we duplicate state, we should either use a middleware or store/restore the local storage through the loading of the network selector in redux or perhaps remove the redux slice for this entierly.

ohenrik (changes requested):

maxwedwards:
I hope you don't mind but I updated the PR with what I think is a more streamlined way of doing this

ziggie1984:
no problem thanks for cleaning this up!

web/src/features/network/networkSlice.ts
```diff
@@ -13,8 +13,18 @@ export interface NetworkState {
...
 }
 
+const defaultNetwork = process.env.NODE_ENV === "development" ? Network.SimNet : Network.MainNet;
+
+const getActiveNetworkFromLocalStorage = (): Network | undefined => {
+  const lsActiveNetwork = localStorage.getItem("activeNetwork");
+  if (!lsActiveNetwork) {
+    return undefined;
+  }
+  return parseInt(lsActiveNetwork) as Network;
```
ziggie1984, line 23: hmm this could lead to a runtime error if the lsActiveNetwork somehow is a string and cannot be parsed as a number?

maxwedwards: no, it just returns NaN

PR #466, opened 2023-02-24
117 closed pending channels table page

laitang2000:


web/src/apiSlice.ts
```diff
@@ -101,6 +101,20 @@ export const torqApi = createApi({
       }),
       providesTags: ["channels"],
     }),
+    getChannelsClosed: builder.query<channel[], ActiveNetwork>({
+      query: (params) => ({
+        url: `channels/closed` + queryParamsBuilder(params),
+        method: "GET",
+      }),
+      providesTags: ["channels"],
```
ohenrik, line 109: This should be something unique. Add `closedChannels` to the tagTypes  array at line 56, then you should replace that here so that `channels` says `closedChannels`. You can then also invalidate that tag when we close channels.

web/src/apiSlice.ts
```diff
@@ -101,6 +101,20 @@ export const torqApi = createApi({
...
+    }),
+    getChannelsPending: builder.query<channel[], ActiveNetwork>({
+      query: (params) => ({
+        url: `channels/pending` + queryParamsBuilder(params),
+        method: "GET",
+      }),
+      providesTags: ["channels"],
```
ohenrik, line 116: same here. `pendigChannels`

web/src/constants/routes.ts
```diff
@@ -11,6 +11,9 @@ export const FORWARDS = "forwards";
 export const TAGS = "tags";
 export const FORWARDS_CUSTOM_VIEW = `${FORWARDS}/:viewId`;
 export const CHANNELS = "channels";
+export const OPEN_CHANNELS = "opened-channels";
```
ohenrik, line 14: Change to `open-channels`

web/src/features/channels/ChannelsPage.tsx
```diff
@@ -33,6 +32,7 @@ import { useGroupBy } from "features/sidebar/sections/group/groupBy";
 import channelsCellRenderer from "./channelsCellRenderer";
 import { selectActiveNetwork } from "features/network/networkSlice";
 import { TableResponses, ViewResponse } from "../viewManagement/types";
+import * as Routes from "../../constants/routes";
```
ohenrik, line 35: relative import

web/src/features/channelsClosed/channelsClosedCellRenderer.tsx
```diff
@@ -0,0 +1,52 @@
...
+    case "tags":
+      return <TagsCell tags={row.tags} key={"tagsCell" + rowIndex} channelId={row.channelId} nodeId={row.peerNodeId} />;
+    case "fundingTransactionHash":
+      return (
+        <LongTextCell
+          current={row.fundingTransactionHash}
+          link={"https://mempool.space/tx/" + row.fundingTransactionHash}
+          copyText={row.fundingTransactionHash}
+        />
+      );
```
ohenrik, line 39: You could add a if statement here to only return this version if the user has also specified the "LongTextCell" cell type. They can select "TextCell" if they want to just see the full string. So add

if (column.type === 'LongTextCell') {
return ....
}
This way the default applies for the other cell types.

web/src/features/channelsClosed/channelsClosedCellRenderer.tsx
```diff
@@ -0,0 +1,52 @@
...
+    case "closingTransactionHash":
+      return (
+        <LongTextCell
+          current={row.closingTransactionHash}
+          link={"https://mempool.space/tx/" + row.closingTransactionHash}
+          copyText={row.closingTransactionHash}
+        />
+      );
+
```
ohenrik, line 48: Same here.

web/src/features/channelsPending/channelsPendingCellRenderer.tsx
```diff
@@ -0,0 +1,52 @@
...
+        />
+      );
+    case "closingTransactionHash":
+      return (
+        <LongTextCell
+          current={row.closingTransactionHash}
+          link={"https://mempool.space/tx/" + row.closingTransactionHash}
+          copyText={row.closingTransactionHash}
+        />
+      );
```
ohenrik, line 47: Same here as well. Conditional so that the user can still select `TextCell` instead of the `LongTextCell` using the UI.

ohenrik (changes requested):
This looks great! I have noted a few things, the most important change is the `providesTags` param that needs to be changed.

PR #473, opened 2023-02-27
Fix update channel policy errors

maxwedwards:
It was a bit of a mess with massive double red Xs and with difficult to understand generic error messages.

Front end code is now using some common error handling code. The style has been updated to use the shared error summary component and the backend has been updated to send specific error messages rather than just always replying with a generic error.


ohenrik (approved):
LGTM. I'm not sure why the codeQL Javascript failed though

PR #483, opened 2023-03-01
Fixed when deleting a node channels will now show channels correctly.

laitang2000:
Backend will reset node and channelstate caches and run through the Initializing process to rehydrate data.

Pending and Closed channels query amended to exclude deleted nodes

### DIFF
--- a/internal/channels/database.go
+++ b/internal/channels/database.go
@@ -301,7 +301,8 @@ func getClosedChannels(db *sqlx.DB, network int) ([]Channel, error) {
 	err := db.Select(&channels, `
 		SELECT c.* FROM Channel c
 		JOIN Node n on n.node_id = c.First_Node_Id
-	 	WHERE n.network =$1 AND c.Status_id in (100,101,102,103,104,105)
+		JOIN Node_Connection_Details ncd on ncd.node_id = n.node_id
+	 	WHERE ncd.status_id != 3 AND n.network =$1 AND c.Status_id in (100,101,102,103,104,105)
 	`, network)
 
 	if err != nil {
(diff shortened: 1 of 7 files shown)


internal/channels/database.go
```diff
@@ -301,7 +301,8 @@ func getClosedChannels(db *sqlx.DB, network int) ([]Channel, error) {
 	err := db.Select(&channels, `
 		SELECT c.* FROM Channel c
 		JOIN Node n on n.node_id = c.First_Node_Id
-	 	WHERE n.network =$1 AND c.Status_id in (100,101,102,103,104,105)
+		JOIN Node_Connection_Details ncd on ncd.node_id = n.node_id
```
2Fast2BCn, line 304: Rather then an extra join you could get the nodeIds from the in memory cache via: GetAllTorqNodeIdsByNetwork and use that in the query to filter.

internal/settings/routes.go
```diff
@@ -491,6 +493,8 @@ func setNodeConnectionDetailsStatusHandler(c *gin.Context, db *sqlx.DB,
 			server_errors.LogAndSendServerError(c, err)
 			return
 		}
+		commons.ResetManagedNodeCache()
+		commons.SendServiceEvent(nodeId, serviceEventChannel, commons.Active, commons.Initializing, commons.TorqService, nil)
```
2Fast2BCn, line 497: maybe we create something else instead of reusing commons.Initializing
I kinda refactored these service statuses so maybe we can do this on top of the rebalancer branch?

maxwedwards:
As per our video call, I would prefer a not so brute force solution. I think we should avoid sending the initialising event and instead remove the node and the node's channels from the various caches.

PR #492, opened 2023-03-07
check for delete status before removing node from cache

laitang2000:


2Fast2BCn:
Looks good. I however don't have permission to merge it :-/

PR #501, opened 2023-03-09
disable enable channel

ziggie1984:
- lnd: fix nil pointer error
- channels: add disable/enable endpoints
- web: add disable/enable button to channelpage


ziggie1984:
This is still prototyping especially the frontend, still not evaluating the response of the button click and aslo adding a normal button instead of a link button.

Also maybe forthwhile to think about also adding the disable/enable button to the `analyse/inspect` modale?

pkg/lnd/lightningCommunication.go
```diff
@@ -220,8 +220,6 @@ func channelStatusUpdateRequestIsRepeated(db *sqlx.DB, request commons.ChannelSt
 			return &commons.ChannelStatusUpdateResponse{
 				Request: request,
 				CommunicationResponse: commons.CommunicationResponse{
-					Status: commons.Inactive,
-					Error:  err.Error(),
```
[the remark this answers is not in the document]
ziggie1984: this is just the ratelimit case, we do not have an error here, err is acutally nil so either we report own error or we return default struct back

ohenrik:
> This is still prototyping especially the frontend, still not evaluating the response of the button click.
I see, no problem.

> Also maybe forthwhile to think about also adding the disable/enable button to the `analyse/inspect` modale?
It would make sense to add it there as well. So please feel free to do so.

web/src/components/table/cells/channelCell/ChannelCell.tsx
```diff
@@ -25,6 +33,69 @@ function ChannelCell(props: ChannelCell) {
...
+
+  const [isDisabled, setDisabled] = useState(false);
+
+  useEffect(() => {
+    if (channelDetail?.active) {
+      setDisabled(false);
+    } else if (!channelDetail?.active) {
+      setDisabled(true);
+    }
+  }, [channelDetail?.active]);
```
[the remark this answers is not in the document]
ziggie1984: makes sense

ziggie1984:
solves lncapital/internal-torq-roadmap#132

web/src/components/buttons/UpdateChanStatusButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const updateChannelState = async () => {
+    await updateChannelStatus(sI);
+    mixpanel.track("Update ChannelStatus", {
+      channelId: props.chanId,
+      nodeId: props.nodeId,
+      status: props.isDisabled ? Status.Active : Status.Inactive,
+    });
+
+    if (isError || updateChannelResponse === undefined) {
```
ziggie1984, line 47: Here is a problem I cannot fix, somehow  when I click very fast the api reports an undefined object here although the backends gets updated. It seems to me that the api somehow ratelimits but the backend is not aware of this and updates the state.

maxwedwards: I will try to recreate and see what's going on

web/src/components/table/cells/alias/AliasCell.tsx
```diff
@@ -23,6 +28,17 @@ interface AliasCell {
...
+  const { data: channelDetails } = useGetChannelsQuery<{
+    data: Array<channel>;
+    isLoading: boolean;
+    isFetching: boolean;
+    isUninitialized: boolean;
+    isSuccess: boolean;
+  }>({ network: activeNetwork }, { pollingInterval: 1000 });
+
+  const channelDetail = channelDetails?.find((c) => c.channelId === channelId);
+
```
maxwedwards, line 41: These cell components should be "dumb" and not do their own web requests. I understand why you are doing it but really we need control on data fetching to be with a parent "smart" component such as the `ChannelsPage.tsx` itself. Ideally, we have the web request be synchronous and we update the state of enabled / disabled when it returns.

web/src/components/table/cells/channelCell/ChannelCell.tsx
```diff
@@ -25,6 +30,17 @@ function ChannelCell(props: ChannelCell) {
...
+  const { data: channelDetails } = useGetChannelsQuery<{
+    data: Array<channel>;
+    isLoading: boolean;
+    isFetching: boolean;
+    isUninitialized: boolean;
+    isSuccess: boolean;
+  }>({ network: activeNetwork }, { pollingInterval: 1000 });
+
+  const channelDetail = channelDetails?.find((c) => c.channelId === props.channelId);
+
```
maxwedwards, line 43: Again, I don't think cell components should be making these requests.

web/src/components/buttons/UpdateChanStatusButton.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+  const toastRef = React.useContext(ToastContext);
+
+  const sI: StatusInterface = {
+    channelId: props.chanId,
+    nodeId: props.nodeId,
+    channelStatus: props.isDisabled ? Status.Active : Status.Inactive,
+  };
+
+  let isDisabled: boolean = props.isDisabled;
```
maxwedwards, line 35: This needs to be a `react.useState` so that the value is preserved through re-renders

web/src/components/buttons/UpdateChanStatusButton.tsx
maxwedwards: I think I triggered it once or twice. Could it be to do with the `LightningCommunication.go`s deduplication?

maxwedwards:
There is a challenge with this ticket with how we display to the user that a) we have requested something and to wait patiently and b) that the thing has happened or failed.

Normally I would suggest we disable button + make spinner until we have a response. But the the LND API I think it's very hard to get a response. Looks like the request we send to LND is asynchronous so we don't know if it succeeds or not.

Currently as this is implemented you can hit the button as many times as you want to   We need to put our heads together on   how the UI should work given these constraints. Tagging @ohenrik and @2Fast2BCn.

maxwedwards:
I've had a chat with @ohenrik and we will try to make some changes to the backend to make the request synchronous so we know that the channel has been successfully disabled / enabled. So we will pick up this PR from here to try and avoid too much back and forth.

Excellent work getting it this far @ziggie1984, your hands were tied with a not easy API to consume!

ziggie1984:
ok thank you for reviewing and I consider my part of it done now :)

PR #502, opened 2023-03-09
Workflow datasource migration.

2Fast2BCn:


2Fast2BCn:
Merge order decides if we need to change the filenames.

ohenrik (changes requested):
Have you manually tested this migration from version 18 to version 19?

PR #503, opened 2023-03-09
Requesting a new set of pending channels from LND.

2Fast2BCn:


2Fast2BCn:
This is now inside PR #504

Issue #469, opened 2023-02-24
enums filter component not working


laitang2000:
When filtering enum on the side panel, selections do not appear

![Image](https://user-images.githubusercontent.com/10725544/221236782-e22e6159-a8f8-42e4-a6ce-d401194437e9.png)

ohenrik:
Scratch that. Needs to be looked at

PR #533, opened 2023-03-15
Fixed enums for payments, invoices and on-chain transactions tables.

ohenrik:
Fixes https://github.com/lncapital/torq/issues/469

NB! We still need to fix the filtering of Arrays and Enums in the processFilter.go file used in the automations!

### DIFF
--- a/internal/workflows/processFilter.go
+++ b/internal/workflows/processFilter.go
@@ -377,6 +377,7 @@ func GetFilterFunctions() map[FilterCategoryType]map[string]FilterFunc {
 			},
 		},
 		FilterCategoryTypeEnum: {
+			// TODO: This needs to be fixed. Enums are using any and notAny instead of like and notLike
 			"like": func(dataMap map[string]interface{}, dataKey string, filterValue FilterParameterType) bool {
 				if isNil(dataMap[dataKey]) != (filterValue == nil) {
 					return false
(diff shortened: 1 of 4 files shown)


internal/workflows/processFilter.go
```diff
@@ -377,6 +377,7 @@ func GetFilterFunctions() map[FilterCategoryType]map[string]FilterFunc {
 			},
 		},
 		FilterCategoryTypeEnum: {
+			// TODO: This needs to be fixed. Enums are using any and notAny instead of like and notLike
```
ohenrik, line 380: This needs to accept an array of values and check if the enum value is equal to any of those chosen in the values.

internal/workflows/processFilter.go
```diff
@@ -691,6 +692,7 @@ func GetFilterFunctions() map[FilterCategoryType]map[string]FilterFunc {
 			},
 		},
 		FilterCategoryTypeArray: {
+			// TODO: This needs to be fixed. Arrays are using any and notAny instead of like and notLike
 			"eq": func(dataMap map[string]interface{}, dataKey string, filterValue FilterParameterType) bool {
 				if isNil(dataMap[dataKey]) != (filterValue == nil) {
 					return false
```
ohenrik, line 698: This needs to accept an array of values and check if the enum value is equal to any of those chosen in the values.

ohenrik (commented):
Enum and Array filtering in the processFilter.go file needs to be fixed

web/src/features/sidebar/sections/filter/FilterRow.tsx
```diff
@@ -182,8 +182,10 @@ function FilterRow({
   const label = filterOptions.find((item) => item.value === rowValues.key)?.label;
   const options = filterOptions.find((item) => item.value === rowValues.key)?.selectOptions || [];
 
-  const getParameter = () => {
+  function getParameter() {
```
maxwedwards, line 185: Why the change in style? I thought all the cool kids liked const funcs and they had better `this` behaviour.

maxwedwards:
Happy for this to be merged when ready

ohenrik:
I'm creating a separate issue for the processFilter.go change needed.

PR #538, opened 2023-03-15
Fix for the pending settlement problem?

2Fast2BCn:


ohenrik (changes requested):
Let's do some more testing on this before we merge this in and let's talk through it first.

PR #189, opened 2022-10-27
#114: Add tags to channels + refactoring (WIP)

2Fast2BCn:

PR #540, opened 2023-03-16
#189: New address buttons should be disabled until you select a node

2Fast2BCn:


web/src/features/transact/newAddress/NewAddressModal.tsx
```diff
@@ -44,6 +45,15 @@ function NewAddressModal() {
...
 
+  useEffect(() => {
+    if (nodeConfigurations) {
+      if (nodeConfigurations.length == 1) {
+        setSelectedNodeId(nodeConfigurationOptions[0].value)
+        setButtonsEnabled(nodeConfigurationOptions[0].value !== 0)
+      }
+    }
+  }, [nodeConfigurations]);
+
```
ohenrik, line 56: Using `useEffect` here is the right direction, but I believe we do not need the `setButtonsEnabled`. We can use `selectedNodeId` combined with `isLoading` from line `useNewAddressMutation` on line 57 to determine if the buttons are disabled or not.

web/src/features/transact/newAddress/NewAddressModal.tsx
```diff
@@ -44,6 +45,15 @@ function NewAddressModal() {
 
   const [selectedNodeId, setSelectedNodeId] = useState<number>(nodeConfigurationOptions[0].value);
```
ohenrik, line 46: I think we should not set the `selectedNodeId` directly here (`const [selectedNodeId, setSelectedNodeId] = useState<number>(nodeConfigurationOptions[0].value);`

Instead we should allow it to be undefined and set it in the useEffect. so:

const [selectedNodeId, setSelectedNodeId] = useState<number | undefined>();`

then:

```
  useEffect(() => {
    if (nodeConfigurations) {
      if (nodeConfigurations.length == 1) {
        setSelectedNodeId(nodeConfigurationOptions[0].value)
      }
    }
  }, [nodeConfigurations]);

File: web/src/features/transact/newAddress/NewAddressModal.tsx [hunk 3]
```diff
@@ -81,7 +94,7 @@ function NewAddressModal() {
           {addressTypeOptions.map((addType, index) => {
             return (
               <Button
-                disabled={isLoading}
+                disabled={isLoading || !buttonsEnabled}
```
ohenrik, line 97: use `disabled={isLoading || !selectedNode}` here.

web/src/features/transact/newAddress/NewAddressModal.tsx
```diff
@@ -69,7 +79,10 @@ function NewAddressModal() {
             label={t.yourNode}
             onChange={(newValue: unknown) => {
               const value = newValue as Option;
-              if (value && value.value > 0) setSelectedNodeId(value.value);
+              if (value && value.value > 0) {
+                setSelectedNodeId(value.value);
+                setButtonsEnabled(true)
+              }
```
ohenrik, line 85: remove `setButtonsEnabled(true)`

web/src/features/transact/newAddress/NewAddressModal.tsx
```diff
@@ -23,6 +23,7 @@ function NewAddressModal() {
   const toastRef = useContext(ToastContext);
 
   const { data: nodeConfigurations } = useGetNodeConfigurationsQuery();
+  const [buttonsEnabled, setButtonsEnabled] = useState(false);
```
ohenrik, line 26: remove

ohenrik (changes requested):

Issue #395, opened 2023-02-05
Embed web/build subfolder via go:embed into the final binary


prusnak:
This allows Torq to be distributed as a single binary simplifying deployment significantly.

ohenrik:
We'll try to add this soon, thank you for adding the issue :) We should probably also create a go binary release as an alternative to the docker image.

prusnak:
> We'll try to add this soon, thank you for adding the issue

Yeah, I remember you mentioning it on our call. I just wanted to make an issue, so this idea is not forgotten.

maxwedwards:
Work is complete, next Torq release will have frontend embedded

prusnak:
cc @mmilata

PR #543, opened 2023-03-17
Go embed

2Fast2BCn:
related to https://github.com/lncapital/torq/issues/395


2Fast2BCn:
@maxwedwards I think the problem here is the token is not able to upload the artifacts to the github releases endpoint. So I was going to try to create a github access token with the correct access and put it in the github action secrets and use that to upload artifacts but I don't have access to those things.

PR #548, opened 2023-03-17
Fix rapid reboot problems.

2Fast2BCn:


maxwedwards:
Massive change, can you give a summary of what and why?

PR #557, opened 2023-03-24
Slack and Telegram support

2Fast2BCn:
fixes https://github.com/lncapital/internal-torq-roadmap/issues/112
fixes https://github.com/lncapital/internal-torq-roadmap/issues/111


database/migrations/000083_notifier.up.psql
```diff
@@ -0,0 +1,19 @@
...
+ALTER TABLE settings ADD COLUMN slack_bot_app_token TEXT;
+ALTER TABLE settings ADD COLUMN telegram_high_priority_credentials TEXT;
+ALTER TABLE settings ADD COLUMN telegram_low_priority_credentials TEXT;
+
+CREATE TABLE communication (
+    communication_id SERIAL PRIMARY KEY,
+    --bitshifted value for the type of communication (new channel, stale fee, ...)
+    communication_type INTEGER NOT NULL,
+    --types like (telegram high, telegram low, email, ...)
+    target_type INTEGER NOT NULL,
```
ohenrik, line 12: Is this really bitshifted? Can it be multiple types or is it more like an enum?

ohenrik: If it's multiple, would it not be better to have each communication type requested to be a separate row? We can then simply loop through the rows instead of exploding the bitshifted value into separate actions in the code, which is a bit tricky to read and will require unit tests to be written.

internal/communications/notifier.go
```diff
@@ -0,0 +1,792 @@
...
+			if communicationMessage != "" {
+				processCommunications(communicationMessage, communications)
+			}
+		}
+	}()
+	wg.Wait()
+	log.Info().Msgf("Notifier initializing NodeGraphEvent terminated")
+}
+
+func notifyOnNotifierEvent(ctx context.Context, db *sqlx.DB) {
```
ohenrik, line 236: Does this function send a notification when requested to do so? If so, then maybe name it `sendNotification`? Easier to reason about and spot when you are looking for that functionality.

internal/communications/notifier.go
```diff
@@ -0,0 +1,792 @@
...
+		messageForBot.Error = err.Error()
+		return messageForBot
+	}
+	for _, communication := range communications {
+		if activate && communication.HasCommunicationType(*communicationType) {
+			continue
+		}
+		if !activate && !communication.HasCommunicationType(*communicationType) {
+			continue
+		}
```
ohenrik, line 572: What does this do? Perhaps add a comment to explain it. To me the first case seems to rule out all active communications.

internal/communications/notifier.go
```diff
@@ -0,0 +1,792 @@
...
+		messageForBot = processUnregisterRequest(db, communicationTargetType, messageForBot)
+	default:
+		messageForBot = processSettingsRequest(db, communicationIds, communicationTargetType, messageFromChannel, messageForBot)
+	}
+	if messageForBot.HasMessage() || messageForBot.HasMenu() {
+		NotificationChannels[communicationTargetType] <- messageForBot
+	}
+}
+
+func processSettingsRequest(db *sqlx.DB,
```
ohenrik, line 485: Is this function used to process communication settings requested by the user? So update / Add / remove?

pkg/lnd/channel_events.go
```diff
@@ -366,12 +366,12 @@ func SubscribeAndStoreChannelEvents(ctx context.Context, client lndClientSubscri
...
-			case <-ctx.Done():
-				return
-			default:
-			}
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		default:
```
ohenrik, line 374: What does this change? Is it a fix for something?

pkg/lnd/channelgraph.go
```diff
@@ -53,12 +53,12 @@ func SubscribeAndStoreChannelGraph(ctx context.Context, client subscribeChannelG
...
-			case <-ctx.Done():
-				return
-			default:
-			}
+		}
+
+		select {
+		case <-ctx.Done():
+			return
+		default:
```
ohenrik, line 61: same here. I assume this fixez some issue?

web/src/apiTypes.ts
```diff
@@ -5,6 +5,10 @@ export interface settings {
   weekStartsOn: "saturday" | "sunday" | "monday";
   torqUuid: string;
   mixpanelOptOut: boolean;
+  slackOAuthToken: string;
+  slackBotAppToken: string;
+  telegramHighPriorityCredentials: string;
+  telegramLowPriorityCredentials: string;
```
ohenrik, line 11: Does the user need to create their own bots? or how do they get these credentials?

web/src/features/settings/SettingsPage.tsx
```diff
@@ -150,6 +167,38 @@ function Settings() {
...
+                  />
+                </span>
+                <span id="telegramLowPriorityCredentials">
+                  <Input
+                    label={t.telegramLowPriorityCredentials}
+                    value={settingsState?.telegramLowPriorityCredentials}
+                    type={"text"}
+                    onChange={(e: React.ChangeEvent<HTMLInputElement>) => handleTelegramLowPriorityCredentialsChange(e.target.value)}
+                  />
+                </span>
```
ohenrik, line 201: These should not to be wrapped in span elements, unless you have a very good reason. the ID can be applied directly on the `Input`

pkg/lnd/channelgraph.go
2Fast2BCn: I assumed you guys were "allergic" to `} else {` statements?

web/src/apiTypes.ts
2Fast2BCn: They need to create their own bots

internal/communications/notifier.go
2Fast2BCn: yes

internal/communications/notifier.go
2Fast2BCn: It isn't actually sending anything yet.
It's monitoring a channel on which events come in from throughout the application and filters and expand those events if needed. It then passes them on to the correct notification platform slack/telegram/... which will actually perform the send operation.

internal/communications/notifier.go
2Fast2BCn: added comments to it

database/migrations/000083_notifier.up.psql
2Fast2BCn: replaced it with a 0/1 field.

internal/settings/routes.go
```diff
@@ -133,26 +140,56 @@ func getTimeZonesHandler(c *gin.Context, db *sqlx.DB) {
...
+	}
+	if startedSomething &&
+		commons.RunningServices[commons.NotifierService].GetStatus(commons.TorqDummyNodeId) == commons.ServiceInactive {
+		serviceChannel <- commons.ServiceChannelMessage{
+			NodeId:         commons.TorqDummyNodeId,
+			ServiceType:    commons.NotifierService,
+			ServiceCommand: commons.Boot,
+		}
+	}
+	c.JSON(http.StatusOK, setts)
```
ohenrik, line 192: Is this something we can replace with the new pattern?

pkg/lnd/channelgraph.go
```diff
@@ -404,33 +408,38 @@ func insertNodeEvent(db *sqlx.DB, eventTime time.Time, eventNodeId int, alias st
...
 		}
+		communications.NotifyOnNotifierEvent(db, commons.NotifierEvent{
+			EventData: commons.EventData{
+				EventTime: eventTime,
+				NodeId:    nodeId,
+			},
+			NotificationType: commons.NodeDetails,
+			NodeGraphEvent:   &nodeGraphEvent,
+		})
 	}
```
ohenrik, line 443: It looks like everything from line 397 (393) and down to here is within an if statement, that can most likely be replaced by an early return or this code should be a separate function.

```
	if alias != nodeEvent.Alias ||
		color != nodeEvent.Color ||
		string(najb) != nodeEvent.NodeAddresses ||
		string(fjb) != nodeEvent.Features {
.....
```

PR #568, opened 2023-04-05
Workflow node deletion via soft delete.

2Fast2BCn:


2Fast2BCn:
Fixes: https://github.com/lncapital/internal-torq-roadmap/issues/209

2Fast2BCn:
I've put the hard delete code in comments because I assumed we would still want it for when we create the database cleanup page.

internal/workflows/database.go
```diff
@@ -967,57 +980,69 @@ func updateNodeVisibilitySettings(db *sqlx.DB, workflowVersionNodeId int, visibi
...
+//		return 0, errors.Wrap(err, "Committing the transaction while deleting the node")
+//	}
+//
+//	_, err = res.RowsAffected()
+//	if err != nil {
+//		return 0, errors.Wrap(err, database.SqlAffectedRowsCheckError)
+//	}
+//
+//	return workflowVersionNodeId, nil
+//}
```
ohenrik, line 1043: Should the be removed?

ohenrik (commented):
I see one problem with this approach. Currently if we delete a node then the name is still in the database which means that if the user want to create another node and save it with the same name it either returns an error or adds a number to it. Should we perhaps add "- deleted" to the name of deleted nodes?

internal/workflows/database.go
2Fast2BCn: > Should the be removed?

I've put the hard delete code in comments because I assumed we would still want it for when we create the database cleanup page.

> I see one problem with this approach. Currently if we delete a node then the name is still in the database which means that if the user want to create another node and save it with the same name it either returns an error or adds a number to it. Should we perhaps add "- deleted" to the name of deleted nodes?

That is exactly what I've done in the PR?

ohenrik: Ok, no problem. Let's keep the commented code. 

And yeah, my bad for not seeing that you handled the duplicate name situation.

PR #581, opened 2023-04-07
Remove fallback for embed.

2Fast2BCn:


ohenrik:
@2Fast2BCn This works locally when visiting :8080, however I think we're not compiling this correctly when we're building the docker image and binaries, since the tests fails. It seems like there is no "build" folder with any content. Could you take a look at the build?

PR #583, opened 2023-04-08
Peers page (rebased)

2Fast2BCn:
https://github.com/lncapital/internal-torq-roadmap/issues/58


ohenrik:
These things need to fix before merging:

* Filtering on enum based peer columns is not working.
* Disconnect and Reconnect does not always work. Seems that either Torq becomes out of sync with the LND connection status or that LND can no longer reconnect.

ohenrik:
Potential problem with disconnection of peer. Not sure if it's worth holding back this feature for release:
https://github.com/lncapital/internal-torq-roadmap/issues/223

PR #588, opened 2023-04-12
Fix connection details singleton when details got updated.

2Fast2BCn:


maxwedwards:
Waiting on update from Kobe to "add close for proper cleanup"

internal/lnd/communication.go
```diff
@@ -32,45 +33,55 @@ var (
...
-//func removeConnection(nodeId int) {
-//	connectionWrapper.mu.Lock()
-//	defer connectionWrapper.mu.Unlock()
-//
-//	_, exists := connectionWrapper.connections[nodeId]
-//	if exists {
-//		delete(connectionWrapper.connections, nodeId)
-//	}
-//}
-
```
maxwedwards, line 73: If we are going to keep commented out code, we should leave a comment to explain why. Would this be useful when someone deletes one of their nodes? Should it also try to call Close() before deleting it from the map?

PR #598, opened 2023-04-14
Channel capacity mismatch using existing cache as fallback.

2Fast2BCn:
https://github.com/lncapital/internal-torq-roadmap/issues/203


2Fast2BCn:
https://github.com/lncapital/internal-torq-roadmap/issues/203

PR #624, opened 2023-04-21
Added functionality to set a peer node as deleted.

laitang2000:
Added functionality to set a peer node as deleted.
Connecting to a previously deleted peer node will change the status to Active.


2Fast2BCn:
Marking a node as deleted or archived feels weird to me.
I could understand archiving the connection history records but why flag the node itself as archived?

I guess my question is what other impact do you want from flagging the node as deleted?

laitang2000:
@2Fast2BCn I'm just reusing the status_id for that we have for nodeConnectionDetails, so the same question can be asked for that, do we do anything special with archived node connection details? If not we can remove that staus

PR #642, opened 2023-04-28
250 - move time interval and add refresh button

laitang2000:
Added TableControlSection to house TimeIntervalSelect and Refresh button for better UX


ohenrik:
Perfect! Great that you added the refresh as well 🎉

Issue #467, opened 2023-02-24
Channels Closed Page - Add Tags column


laitang2000:
On the initial creation of the ChannelsClosedPage Tags column was created but did not function correctly and was removed to avoid delaying original ticket.

Jhoyola:
I can do this. Was the problem just that the tags couldn't be seen or is there something more complex involved? I assume that the tags themselves should not change in anyway when closing a channel.

PR #654, opened 2023-05-03
Add tags column to closed channels and pending channels pages

Jhoyola:
This issue: https://github.com/lncapital/torq/issues/467

Added the Tags column as the third column by default on Pending and Closed channels pages.

### DIFF
--- a/internal/views/views.go
+++ b/internal/views/views.go
@@ -324,9 +324,11 @@ func getTableViewColumnDefinitions() []tableViewColumnDefinition {
 			visualType: "TagsCell",
 			valueType:  "tag",
 			pages: map[TableViewPage]int{
-				PageChannels: 4,
-				PageForwards: 7,
-				PagePeers:    6,
+				PageChannels:        4,
+				PageChannelsClosed:  19,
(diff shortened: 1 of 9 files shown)


internal/views/views.go
```diff
@@ -324,9 +324,11 @@ func getTableViewColumnDefinitions() []tableViewColumnDefinition {
 			visualType: "TagsCell",
 			valueType:  "tag",
 			pages: map[TableViewPage]int{
-				PageChannels: 4,
-				PageForwards: 7,
-				PagePeers:    6,
+				PageChannels:        4,
+				PageChannelsClosed:  19,
```
Jhoyola, line 328: Didn't fully understand these mapping numbers. I guess it doesn't matter what they are as long as they are unique?

2Fast2BCn: They indicate the sequence. So for `PageChannels: 4,` it's the 4th on the channels page.

maxwedwards:
This is quite impressive @Jhoyola as this area of the code is a bit convoluted.

Thanks for the PR!

PR #663, opened 2023-05-09
Added node colour setting to node settings

laitang2000:
Added node colour setting to node settings
Display node colour for channels using override on row first cell.

Fixes: https://github.com/lncapital/internal-torq-roadmap/issues/140
Fixes https://github.com/lncapital/internal-torq-roadmap/issues/165


ohenrik (approved):
All good! 

(Ps. I fixed the peers alias cell I broke earlier)

